Issue #33, opened 2016-08-07
UtkScan Compile Error output_his


ksmith0:
After commenting out utkscan files from issue #31 and switching the flag USE_GSL=OFF to avoid issue #32 the following compile errors were encountered. USE_POLL=OFF

```
[ 93%] Linking CXX executable utkscan
/usr/lib/gcc/x86_64-redhat-linux/5.3.1/../../../../lib64/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
core/CMakeFiles/CoreObjects.dir/HisFile.cpp.o: In function `hd1d_(int, int, int, int, int, int, char const*, unsigned int)':
HisFile.cpp:(.text+0x5e74): undefined reference to `output_his'
HisFile.cpp:(.text+0x5ec7): undefined reference to `output_his'
core/CMakeFiles/CoreObjects.dir/HisFile.cpp.o: In function `hd2d_(int, int, int, int, int, int, int, int, int, int, char const*, unsigned int)':
HisFile.cpp:(.text+0x5f39): undefined reference to `output_his'
HisFile.cpp:(.text+0x5fa8): undefined reference to `output_his'
core/CMakeFiles/CoreObjects.dir/HisFile.cpp.o: In function `count1cc_(int const&, int const&, int const&)':
HisFile.cpp:(.text+0x6b53): undefined reference to `output_his'
core/CMakeFiles/CoreObjects.dir/HisFile.cpp.o:HisFile.cpp:(.text+0x6b83): more undefined references to `output_his' follow
collect2: error: ld returned 1 exit status
Scan/UtkScan/source/CMakeFiles/utkscan.dir/build.make:172: recipe for target 'Scan/UtkScan/source/utkscan' failed
make[2]: *** [Scan/UtkScan/source/utkscan] Error 1
CMakeFiles/Makefile2:450: recipe for target 'Scan/UtkScan/source/CMakeFiles/utkscan.dir/all' failed
make[1]: *** [Scan/UtkScan/source/CMakeFiles/utkscan.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

spaulaus:
All of these errors make me feel like something got garbled in a merge. I will look into it after finishing with the overhaul to the Unpacker/ScanInterface (issue #29 )

ksmith0:
[...]

Issue #31, opened 2016-08-07
UtkScan Compile Error missing source code


ksmith0:
Fresh compile of paass on master and dev branch with USE_POLL=OFF

```
CMake Error at Scan/source/core/CMakeLists.txt:40 (add_library):
 Cannot find source file:
../utkscan/utkscan.cpp
```

ksmith0:
I cannot find this file utkscan.cpp

ksmith0:
USE_SCAN and USE_SCAN_LIBS have no effect, BUILD_UTKSCAN=OFF removes the error.

PR #35, opened 2016-08-07
Merge patch for compile issues.

ksmith0:
A patch for issues #31 and #33 was committed to dev and should be merged with master so users have a stable code base.


spaulaus:
Agreed.

PR #80, opened 2016-09-16
Merge Dev into Master

spaulaus:
Now that we have settled on a reasonable commit scheme, I believe we should start implementing it. We have used many of the updates in the dev branch for quite a while.


ksmith0:
According the scheme this merge should be tested for stability. @spaulaus can you confirm this merge is ok for utkscan? @cthornsb do you have any concerns? I have a couple notes of items I wanted to look at first. I'll look tomorrow and post my results.

ksmith0 (changes requested):
Found a bug (#84) that will cause issues on some machines and I think it should be resolved before release.

spaulaus (commented):
The issue that @ksmith0 mentioned as problematic [#84](https://github.com/pixie16/paass/issues/84) is an issue with the PLX and XIA APIs. I do not think that we need to fix this issue before pushing to master since we have no control over it.

ksmith0:
Issue #86 is a more serious issue that should be resolved.

spaulaus:
@ksmith0 : I agree.

ksmith0 (approved):

PR #88, opened 2016-09-19
Removed CommandString, replaced with std::string.

ksmith0:
This merge removes the CommandString class and corrects a bug with the cursor wrapping from left to right.


spaulaus (approved):
I do not see any major issues with this commit. It looks like it simplifies things by removing an unnecessary class.

PR #90, opened 2016-09-19
Added compiling of PxiDump back to CMake.

ksmith0:


spaulaus (approved):
I see no issue with this change.

PR #97, opened 2016-09-20
Added check for back tab (Shift tab).

ksmith0:
The terminal now ignores this keypress.


spaulaus (approved):
This commit resolves the issue with the Shift+tab inserting an "a" into the terminal.

Issue #94, opened 2016-09-20
Scope fitting too wide.


ksmith0:
@wapmanzoo reports that the scope fitting appears to have limits which make the fit too wide for some very narrow signals.

ksmith0:
The PaulauskasFunc defined as

``` c++
double PaulauskasFitFunc(double *x, double *p) {
    float diff = (x[0] - p[2])/ADC_TIME_STEP;
    if (diff < 0 ) return 0;
    return p[0] + p[1] * exp(-diff * p[3]) * (1 - exp(-pow(diff * p[4],4)));
}
```

 should not return 0 when diff is less than 0, but the baseline p[0].

``` c++
    if (diff < 0 ) return p[0];
```

In addition, the limits on the function should be inspected. In principal no limits should be specified.

PR #98, opened 2016-09-20
Corrected scope fitting to resolve issue #94.

ksmith0:
The fitting function was mistreating the portion when diff was less than
zero. This should have returned the baseline.

In addition, removed any parameter limits and added documentation for
the function.


wapmanzoo (commented):
Making a comment to start the review. Now trying to find how to pull the code to test it.

spaulaus (approved):
Now that the baseline is fixed, fit results are identical to other software.

Issue #86, opened 2016-09-17
Segmentation Fault when stopping run


spaulaus:
I stopped an ldf run and received a segmentation fault. You can see the sequence of commands in the figure below. I believe @wapmanzoo has run into this issue before, but I cannot find an issue log for it. 
![image](https://cloud.githubusercontent.com/assets/1553042/18609746/dc8587c0-7cd7-11e6-8a08-8a94c0330748.png)

wapmanzoo:
Don't remember the specifics. The pixie16msg might have more info. If you have trouble reading the ldf then use lemor and then 

> ouf name.ldf
> That will close it so if is readable.

-Bill

William A. Peters Ph.D.
ORNL
Univ. of Tenn. 
865 576-6798

spaulaus:
I have no trouble reading the ldf or creating histograms.

spaulaus:
Had another segfault, and then could not start poll2 with the `-f` flag. Here is the relevant portion of the Pixie16Msg.txt

![image](https://cloud.githubusercontent.com/assets/1553042/18610393/52272a22-7ce9-11e6-8554-728eea590962.png)

spaulaus:
In a continuation of this issue, poll2 has frozen while stopping a run. I cannot use ctrl+c to kill, ctrl+z to suspend or any other fancy key combination to recover the terminal.

ksmith0:
`Ctrl+C` and `Ctrl+Z` are trapped by the terminal. I believe this issue is caused by the run file being closed prior to the run being ended. Introduced in commit ee8c03f4bb21160699c8625d7c98aed4367e7ef4.

ksmith0:
Should be resolved in commit b17d0f0. Closing this issue until it reappears.

Issue #50, opened 2016-08-25
Poll2 continues to run after disk is full


cthornsb:
Poll2 continues to run after the disk fills up. It displays an error message but continues to run the acquisition with an [ERROR] flag in the status bar. If data cannot be written to disk, then the acquisition probably doesn't need to be running. This is an issue on un-monitored overnight runs for example.

ksmith0:
This is an extension of pixie16/PixieSuite2#47 (Log file stops working if disk is full)

ksmith0:
This issue was resolved in b17d0f0

PR #100, opened 2016-09-21
Corrected issue with closing files.

ksmith0:
The run files are now closed in the RunControl thread to avoid race
conditions. Updated the messages about opening and closing files to use
leader prints. Moved opening a file to RunControl to avoid any future
race conditions.

Changed the Close() method to run CloseOutputFile() only if
a file is open.

This resolves issue numbers #50 and #86.


ksmith0:
Squash and merge please

spaulaus (approved):
Extensively tested on daq-kyle. I believe this should be working now.

PR #102, opened 2016-09-22
Removed poll2 clo cmd, used testing commit 63e13e

ksmith0:
The clo command was added only for making tests and should not have been committed.


spaulaus (approved):
I do not see an issue here. It should not affect any other pieces of the code.

Issue #70, opened 2016-09-03
Backspace does not delete characters


spaulaus:
If I type something wrong in the prompt, I cannot backspace to delete the offending character, the backspace actually pushes me forward.

spaulaus:
I just leaned that I can use the arrow keys and the delete key to emulate the backspace functionality.

ksmith0:
Can you tell me more about when this occurred?

Perhaps try the new `CTerminalTest` program by building with the CMake option `BUILD_TESTS`

The launch CTerminalTest and type `debug` then provide the result after hitting the backspace key. Should look like this:

```
This is test terminal.
Test> debug
Debug mode is on.
TERM: Curs (6, 43) Key: 263 ^G
Test> 
```

ksmith0:
Closing as the issue is not repeatable.

spaulaus:
This issue is repeatable. I've been able to do it twice today. 

The terminal debugger provides 
For backspace key : `TERM: Curs (6-6, 22) Key: 8`
For Delete key : `TERM: Curs (7-6, 22) Key: 330 J`

ksmith0:
From ncurses.h:
```c
#define KEY_BACKSPACE   0407     /* backspace key */
...
#define KEY_DC    0512     /* delete-character key */
```
Octal code 0407 = 263 and 0512 = 220.

The odd thing is that your backspace key gives a code of 8. The code sent is dependent on the terminal used. So if you could provide which terminal you are using that would be helpful as well.

Can you provide the output of `stty -a`?
```
$ stty -a
speed 38400 baud; rows 46; columns 170; line = 0;
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W;
lnext = ^V; flush = ^O; min = 1; time = 0;
-parenb -parodd -cmspar cs8 -hupcl -cstopb cread -clocal -crtscts
-ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl ixon -ixoff -iuclc -ixany -imaxbel -iutf8
opost -olcuc -ocrnl onlcr -onocr -onlret -ofill -ofdel nl0 cr0 tab0 bs0 vt0 ff0
isig icanon iexten echo echoe echok -echonl -noflsh -xcase -tostop -echoprt echoctl echoke
```

[...]

PR #104, opened 2016-09-22
Organization of utkscan header files

spaulaus:
These headers were making the directory unreadable and impossible quickly find information. The files have been reorganized and the source files updated to reflect this change.


ksmith0:
This isn't what I expected. I would have gone the other way and had the following:
utkscan
|-analyzers
| |-include
| |-source
|-core
| |-include
| |-source
...

ksmith0:
Also, I dislike putting paths in the include statements as the movement of one header file requires touching everything that uses it. I prefer to let the compiler figure out where the files are.

spaulaus:
Fair point (though moot with IDEs.....). I like the suggestion and will make the correction.

spaulaus:
@ksmith0 How does this strike your fancy?

Scan/utkscan/CMakeLists.txt
```diff
@@ -34,16 +30,65 @@ if(UTKSCAN_VERBOSE)
...
+#------------------------------------------------------------------------------
+
+#Add the includes from ScanLib to our tree
+include_directories(include ../ScanLib/include)
+
+#Add the local include directories to the build tree
+include_directories(include analyzers/include)
+include_directories(include core/include)
+include_directories(include experiment/include)
+include_directories(include processors/include)
```
ksmith0: Why do you repeatedly add the include directory?
Does all the following source need all these includes?

ksmith0 (changes requested):
I would have expected something like

``` cmake
#All utkscan source needs the following headers
include_directories(core/include)

add_subdirectory(core/source)

add_subdirectory(analyzers)
add_subdirectory(processors)
add_subdirectory(experiments)
```

And the CMakeLists.txt in analyzers etc would have

``` cmake
include_directories(include)
add_subdirectory(source)
```

Scan/utkscan/CMakeLists.txt
spaulaus: Not all of it, but it's such a mixed bag that it's easier than cherry picking which ones need it and which do not. Remember that we have a mess of circular dependencies here.

Scan/utkscan/CMakeLists.txt
```diff
@@ -34,16 +30,65 @@ if(UTKSCAN_VERBOSE)
...
 #Check if GSL is installed
 if(USE_GSL)
     find_package(GSL REQUIRED)
     add_definitions("-D usegsl")
 endif(USE_GSL)
 
+#------------------------------------------------------------------------------
+
+#Add the includes from ScanLib to our tree
+include_directories(include ../ScanLib/include)
```
ksmith0: This line shouldn't be here. the ScanLib/include is added in the parent CMakeLists.txt

Scan/utkscan/CMakeLists.txt
ksmith0: Ok I'll accept that, but you need to drop the repeated addition of the include folder.

Scan/utkscan/CMakeLists.txt
spaulaus: When removing this line, I get a compilation error 

```
[ 84%] Building CXX object Scan/utkscan/core/source/CMakeFiles/CoreObjects.dir/utkscan.cpp.o
In file included from /home/vincent/programs/sources/utk/paass/Scan/utkscan/core/include/UtkScanInterface.hpp:7:0,
                 from /home/vincent/programs/sources/utk/paass/Scan/utkscan/core/source/utkscan.cpp:4:
/opt/PixieSuite2/include/ScanInterface.hpp:127:24: error: ‘XiaData’ has not been declared
  virtual bool AddEvent(XiaData *event_){ return false; }
                        ^~~~~~~
make[3]: *** [Scan/utkscan/core/source/CMakeFiles/CoreObjects.dir/build.make:447: Scan/utkscan/core/source/CMakeFiles/CoreObjects.dir/utkscan.cpp.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/Makefile2:657: Scan/utkscan/core/source/CMakeFiles/CoreObjects.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
[ 84%] Built target ProcessorObjects
make[1]: *** [CMakeFiles/Makefile2:559: Scan/utkscan/CMakeFiles/utkscan.dir/rule] Error 2
make: *** [Makefile:266: utkscan] Error 2
```

ksmith0:
I clearly do not know how to use this review process yet. The requested changes are removal of 

``` cmake
include_directories(include ../ScanLib/include)
```

and remove repeated `include` from

``` cmake
include_directories(include analyzers/include)
include_directories(include core/include)
include_directories(include experiment/include)
include_directories(include processors/include)
```

spaulaus:
I think we did exactly the same things....

ksmith0:
I see you resolved the `BUILD_UTKSCAN_TESTS` issue. I propose that before we merge this we (you) move the tests folder to live inside analyzers. Either analyzers/tests or analyzers/source/tests

ksmith0 (changes requested):
Changes to include_directories is good. Propose we also move tests while we are at it, I imagine the same thing will happen with them in the future.

spaulaus:
I can see some merit in this. Though, each of the directories will get their own tests folders eventually. I was trying to centrally locate the tests, but as you say with the includes/sources it may make more sense to colocate all of them together.

ksmith0:
I agree that each one will eventually get a test folder. This is also already true for `Core/tests`. Is the plan to move tests or not?

spaulaus:
None of the core files have tests yet. I apologize for not being clear. I agree with the move of tests from utkscan/tests to utkscan/analyzers/tests.

ksmith0:
Yes they [do](https://github.com/pixie16/paass/tree/master/Core/tests), although it is more of a test program. (Some info about use in issue #70)

spaulaus:
Oh sorry! We have two "Cores" there's a core for utkscan as well, which is what I assumed was the reference given that this thread is mostly about that!

ksmith0:
Ahh yeah, that is confusing.

spaulaus:
@ksmith0 The request changes have been made. The organization is a little tighter, and more expansible.

ksmith0 (approved):
Looks good. Squash and merge?

spaulaus:
Do it.

PR #110, opened 2016-09-28
Fixed bug in BarDetector::GetCorTimeAve()

spaulaus:
This bug was identified in pixie16/pixie_scan/#26


ksmith0 (approved):
Looks like the correct way to compute an average.

Issue #24, opened 2016-08-03
SHM mode Crashes computer


spaulaus:
The SHM mode has a tendency to crash the computer at high data rates. This was experienced at the recent ND experiment. @cthornsb @ksmith0, can you comment more on this issue?

ksmith0:
I don't have any specific numbers, but we believe it overwhelms the network interface. The solution here is to move to real shared memory.

ksmith0:
@cthornsb as a stop gap can we introduce a downscale for info sent to shm?

PR #111, opened 2016-09-28
Fixes a memory leak in utkscan

spaulaus:
This memory leak causes the software to use all of the available memory and eventually crashes the computer. This may be related to issue #24, but that is unconfirmed at this point.

### DIFF
--- a/Scan/utkscan/core/source/RawEvent.cpp
+++ b/Scan/utkscan/core/source/RawEvent.cpp
@@ -31,6 +31,10 @@ void RawEvent::Zero(const std::set<std::string> &usedev) {
         (*it).second.Zero();
     }
 
+    for(vector<ChanEvent*>::iterator it = eventList.begin();
+                it != eventList.end(); it++)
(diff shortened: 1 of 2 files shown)


ksmith0 (approved):
Seems like a reasonable commit. Int he future would try to avoid committing files where only white space was touched. I tested in on my laptop and observed no leak in memory scanning a file although no damm or any output was produced so I could not check that the histograms were reasonable.

ksmith0:
Ahh, the arguments are different for utkscan I see. I need to add `-o output` to produce the DAMM output. I seem to be missing half of the raw spectrum, the DAMM output only goes to Looks like damn only plot 15 bit output instead of the full 16 bit.

spaulaus:
Ah, sorry about that. I had put in a debugging statement there and didn't cleanly remove it. 

DAMM is not capable of producing anything larger than a 14 bit histogram. In addition, the control of the raw histograms is determined by the xml file that you are using to test.

PR #112, opened 2016-09-29
Refactoring UtkUnpacker to process scalars and histograms

spaulaus:
UtkUnpacker has been refactored to now plot some useful histograms. It also plots scalars for the different channels and provides us with diagnostic outputs and status updates. Most importantly we now have information about the run time and the total number of counts we had per second and per millisecond.

### DIFF
--- a/Scan/ScanLib/include/XiaData.hpp
+++ b/Scan/ScanLib/include/XiaData.hpp
@@ -51,7 +51,7 @@ public:
     virtual ~XiaData();
     
     /// Get the event ID number (mod * chan).
-    int getID(){ return(modNum*16+chanNum); }
+    unsigned int getID(){ return(modNum*16+chanNum); }
     
     /// Reserve specified number of bins for the channel trace.
     void reserve(const size_t &size_);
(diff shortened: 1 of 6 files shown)


spaulaus:
@ksmith0 I am once again unable to reproduce the behavior seen on your laptop. We should confirm the cmake cache. 

Its possible that this is due to a misconfigured XML file (there's not a lot of error checking for that unfortunately). There is a configuration file included with this branch called `newhist.xml`. I also have a PLD file that I can pass along that works with that XML, or edit the XML to suit your needs.

ksmith0:
You are referring to the large increase in load and large memory usage (6 GB)? I am making a new DAMM histogram if that makes a difference. The cmake flag is `USE_HRIBF=OFF`

Here is my [CMakeCache.txt](https://github.com/pixie16/paass/files/520011/CMakeCache.txt) and the config file I used [config.xml.txt](https://github.com/pixie16/paass/files/520017/config.xml.txt).

ksmith0:
My earlier issue has been spun off into issue #121. It is a "misconfiguration" in the config file. I will now test this pull request with a working config file.

ksmith0:
The damm histograms were not filled for a set of CeBr3 test files used.

The following message was displayed, but it seems to have no bearing on data output

```
ReadSpill: Received spill of 65569 words, but read 65567 words
ReadSpill: Spill split between buffers 
```

Using `--counts` I see that there were triggered data in M0C0 and M0C1 detected by the unpacker as expected.

spaulaus:
Can you share that file with me so that I can test this? I am in the process of uploading my test data files to OneDrive for ease of sharing.

ksmith0 (approved):
Works as expected.

Issue #113, opened 2016-09-29
PLD_header::run_title not initialized causes seg fault


ksmith0:
The `PLD_header` destructor calls 

``` c++
PLD_header::~PLD_header(){                                                           
   if(run_title){ delete[] run_title; }                                              
}  
```

`run_title` is only initialized in `PLD_header::Reset()`. Need to initialize or call Reset() in constructor.

ksmith0:
A fix was committed upstream.

PR #114, opened 2016-09-29
Resolving issue #113 PLD_header seg fault.

ksmith0:
Added PLD_header::Reset() to PLD_header constructor.


spaulaus (approved):
Everything builds without issue. I see no problems with the addition.

PR #116, opened 2016-09-30
Critical patches

ksmith0:


spaulaus (approved):
I do not see any reason not to merge these immediately. We have tested and verified that they all work in their independent reviews.

PR #123, opened 2016-10-12
Changed UtkUnpacker call to destructor to delete.

ksmith0:
We try to avoid direct calls to the destructor.


spaulaus (approved):
I have no issues with this change. It is better practice and works fine.

PR #124, opened 2016-10-12
Updated doxygen configuration file.

ksmith0:
The default configuration now adds the source browser and use dot to
generate graphs including the caller graph.


spaulaus (approved):
These graphs and the source browser can be incredibly useful when trying to understand the software. It does what it says it does.

PR #125, opened 2016-10-14
Corrected utkscan cmake install path for share.

ksmith0:


spaulaus (approved):
This does not cause any issues. It removes a duplicated character in the installation path.

PR #126, opened 2016-10-20
pread/write uses ranges and reads hex input.

ksmith0:
Added the capability to specify modules or channels with a colon character delimiting a range. For example,
`pwrite 0 0:4 TRIGGER_THRESHOLD 0x1a`
would set the trigger threshold for module 0 and channels 0-4 to the hex value 0x1a (decimal 26).

### DIFF
--- a/Poll/include/poll2_core.h
+++ b/Poll/include/poll2_core.h
@@ -235,6 +235,13 @@ class Poll{
 	/// Broadcast a data spill onto the network in the classic pacman format.
 	void broadcast_pac_data();
 
+	/// @brief Splits the arguments to pread and pwrite on a colon delimeter.
+	/// @param[in] arg The argument to be split.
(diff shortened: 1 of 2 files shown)


spaulaus (commented):
@sztaylor89 and I have tested this branch. The branch works as described but is missing useful features. @ksmith0  can you comment on these?

1. Colon range does not work with adjust_offsets, pmread/write, toggle, get_traces
2. csr_test does not accept hex input. 

Was there not time to implement this, or are these more difficult to update?

ksmith0:
These things can be fairly easily modified. I agree that the changes should be consistent throughout as that is likely what the user would expect.

spaulaus:
I will approve this request based on the fact that it works as intended. I will also open a new issue reminding us to update the rest.

spaulaus (approved):
Works as intended.

PR #127, opened 2016-10-26
Parse debug

ksmith0:
Changes to output message when parsing fails. These are now more useful by providing the last event, the current event and the next event.

### DIFF
--- a/Poll/source/poll2_core.cpp
+++ b/Poll/source/poll2_core.cpp
@@ -1580,7 +1580,7 @@ void Poll::RunControl(){
 			//Handle a stop signal
 			if(do_stop_acq){ 
 				// Read data from the modules.
-				ReadFIFO();
+				if (!had_error) ReadFIFO();
 
 				// Instruct all modules to end the current run.
 				pif->EndRun();
@@ -1596,7 +1596,7 @@ void Poll::RunControl(){
 						//We sleep to allow the module to finish.
 						sleep(1);
 						//We read the FIFO out.
-						ReadFIFO();
+						if (!had_error) ReadFIFO();
 					}
 
 					//Print the module status.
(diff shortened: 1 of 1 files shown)


spaulaus:
How do we propose to test this? The errors for the parsing can be difficult to replicate.

ksmith0:
Try `pmwrite 0 SlotID 3`

spaulaus (commented):
@sztaylor89 and I have tested this branch. I have attached pictures of the output. The information is much more useful than it once was. However we did notice some interesting features. 

## Testing Procedure
1. Setup acquisition to verify that we are taking data properly. 
2. Used montior to verify that the data rates were correct. 
3. Used mca to verify that the spectra were appropriate. 
1. Change SlotID : `pmwrite 0 SlotID 5`
2. `startvme` -> Now we have received the error message: 
https://drive.google.com/file/d/0B_XQp_kYuMwQZGo2S2drc0pLYkU/view?usp=sharing
7. Change back to SlotID 2. 
8. `startvme` -> This still produces the same errors as before. 
https://drive.google.com/file/d/0B_XQp_kYuMwQZ2xPMmFta0FacEk/view?usp=sharing
8. Restart poll 
9. Enable traces on Module 0 Channel 0. 
10. Change SlotID to 5
11. `startvme` -> https://drive.google.com/file/d/0B_XQp_kYuMwQSlQ3WmFkX1BVNGs/view?usp=sharing
12. Change SlotID to 2
13. `startvme` -> The system runs as expected with no issues. 

# Summary 
It seems to me that there is something deeper going on when we are modifying the SlotID. Could this be something with the API? It would also seem that poll2 is able to handle recovery after reading traces much better than when it just has the four header words. I think that this needs to be investigated a little more.

ksmith0:
I believe that the modules were not left in a clean state. I'm guessing the `PixieInterface::EndRun` was either not called or failed. This issue is a deeper issue and is not directly related to this pull request, as it was intended to update the debug output. I'd recommend we move your notes to a new issue and approve this request as it intention worked as expected.

spaulaus (approved):
This works as intended.

PR #128, opened 2016-10-26
Multiconfig

ksmith0:
Added ability to boot an arbitrary number of different module types. The cmake package FindPXI now scans for multiple firmwares and builds a pixie.cfg file with every type found.


spaulaus (dismissed):
We did a relatively extensive review with this branch yesterday. Overall things were positive. We did however find an issue that requires revision. 

# Issues that require attention
1. When the pixie.cfg is generated by cmake it may not be able to parse the proper module type associated with the firmware from the name of the files. In our case we had a Rev D installation with a folder name of `/opt/xia/pixie16_revd_general_release_08292015/firmware/` and firmware files named like `syspixie16_revdgeneral_r33157.bin`. This caused the autogenerated pixie.cfg to create an entry labeled as `unknown-revd`. During boot this caused the system to crash and the computer required a hard reset. Amending this line to the appropriate module type (in this case `12b100m-revd`) fixed all of the issues. 

# Issues that we would like fixed but are not critical
1. The comment line reading `Multi / zero configs...` that appears in the pixie.cfg should be amended to be clearer. 
2. If we remove the ModuleBaseDir from a configuration block we should have that value replaced with the default path and this should be output when the configuration information is printed during boot. 
3. Information about the path to the crate configuration (`pxisys.ini`) should be added to the information output during the booting procedure.
4. Sorting pixie.cfg by revision (?) This may not be critical but would be nice. 
5. If a duplicate entry for a module is found in the pixie.cfg, the boot process should abort. The program should tell the user of the issue and tell them that this will need to be fixed (comment or remove duplicates). We have to assume that the user will know enough about the system to be editing this file. Especially, if running in such a non-standard mode.

ksmith0:
>1. The comment line reading Multi / zero configs... that appears in the pixie.cfg should be amended to be clearer.

This was amended in commit b880905.

>1. When the pixie.cfg is generated by cmake it may not be able to parse the proper module type associated with the firmware from the name of the files. In our case we had a Rev D installation with a folder name of /opt/xia/pixie16_revd_general_release_08292015/firmware/ and firmware files named like syspixie16_revdgeneral_r33157.bin. This caused the autogenerated pixie.cfg to create an entry labeled as unknown-revd. During boot this caused the system to crash and the computer required a hard reset. Amending this line to the appropriate module type (in this case 12b100m-revd) fixed all of the issues.

Tested with `ModuleType unknown-revd` on r230 and found the following error message and it did not crash.

```
[ERROR] Invalid ModuleType, ADC resolution not specified: 'unknown-revd'.
[ERROR] Invalid ModuleType, sample rate not specified: 'unknown-revd'.
Module Type: invalid
 ModuleBaseDir  /opt/xia/pixie16_revd_general_release_08292015
 SpFpgaFile     firmware/fippixie16_revdgeneral_r32495.bin
 ComFpgaFile    firmware/syspixie16_revdgeneral_r33157.bin
 DspConfFile    dsp/Pixie16DSP_revdgeneral_r32777.ldr
 DspVarFile     dsp/Pixie16DSP_revdgeneral_r32777.var

...

[ERROR] Configuration not defined for type 12b100m-revd (mod 2)
```

ksmith0:
> 4. Sorting pixie.cfg by revision (?) This may not be critical but would be nice.

This is more work as the `ModuleType` is determined after entering a directory and the output is written out while parsing the files inside that directory. To do as you ask one would need to store the output into a string (can you store multiple lines in a cmake string?). This string is somehow associated with the tag then you would need to sort the tags (not sure this is possible in cmake) and output the associated string. 

I'd think this may be an enhancement issue request.

ksmith0:
>3. Information about the path to the crate configuration (pxisys.ini) should be added to the information output during the booting procedure.

>5. If a duplicate entry for a module is found in the pixie.cfg, the boot process should abort. The program should tell the user of the issue and tell them that this will need to be fixed (comment or remove duplicates). We have to assume that the user will know enough about the system to be editing this file. Especially, if running in such a non-standard mode.

Commit 7c3b387 addressed the above concerns.

ksmith0:
>2. If we remove the ModuleBaseDir from a configuration block we should have that value replaced with the default path and this should be output when the configuration information is printed during boot.

Addressed in commit 08c57d8.

spaulaus (approved):
All of the addressed points work as expected except for 5. The duplicate entry is not ignored and the booting procedure continues as it was before. I do not see anything that should keep this from being pushed.

ksmith0:
First, I did nothing to address point 1 of "Issues that require attention". I simply reported what happened on r230. I'm surprised you went ahead and merged this.

Secondly, I have the following behavior when testing point 5 on r230:
```
Module Type: 12b250m-revf
 PixieBaseDir   /opt/xia/2016-02-02-12b250m-vandle
 SpFpgaFile     firmware/fippixie16_revfvandle_12b250m_r34302.bin
 ComFpgaFile    firmware/syspixie16_revfvandle_adc250mhz_r34301.bin
 DspConfFile    dsp/Pixie16DSP_revfvandle_12b250m_r34300.ldr
 DspVarFile     dsp/Pixie16DSP_revfvandle_12b250m_r34300.var
Module Type: 12b250m-revf
[ERROR] Duplicate module type information found for 12b250m-revf!
        Remove or comment out tags to be ignored.
 ModuleBaseDir  /opt/xia/pixie16_revf_12b250m_vandle_10142016
 SpFpgaFile     firmware/fippixie16_revfvandle_12b250m_r34302.bin
 ComFpgaFile    firmware/syspixie16_revfvandle_adc250mhz_r34301.bin
 DspConfFile    dsp/Pixie16DSP_revfvandle_12b250m_r36100.ldr
 DspVarFile     dsp/Pixie16DSP_revfvandle_12b250m_r36100.var
...
[ERROR] Unable to read configuration file: 'pixie.cfg'
```

spaulaus:
You did address that point by causing the code to catch that the unknown revision and aborting the booting process. This neatly side steps the issue of locking up the computer. At this point it falls back on the user to know what firmware they need and what module they have. I think that's reasonable in this type of advanced configuration.

ksmith0:
Sorry I was not clear enough. Using the same commit as you had when the computer hard locked I had the described behavior of catching the unknown tag and exiting gracefully. I was not able to duplicate the behavior you reported and made no attempt to resolve it. The gracefully exiting code has been part of the branch prior to the review.

spaulaus:
Then we should open an issue for that particular instance. The way I see it that wasn't part of this test. It was related and brought the bug to light, but it wasn't a bug in how the changes for the multi-config system performed.

ksmith0:
I agree that an issue now needs to be opened, but I disagree about this not being a bug from this branch. The `ModuleType` tag did not exist before this branch and thus was introduced by the work done here.

PR #129, opened 2016-10-27
Removed PLD_header delete statement

ksmith0:
PLD_header::~PLD_header had an uneeded delete statement. This was causing a seg fault when quitting poll2 on nitemp.


spaulaus:
This patch will need to be tested on nitemp. Is this the in-crate computer of @wapman's?

ksmith0:
I have tested it on nitemp and it works, I did the development work from there. nitemp is @wapmanzoo's in crate computer.

spaulaus (approved):
Taking @ksmith0 's word that it has been tested and works. We have not had any further user complaints about the issue.

PR #132, opened 2016-11-29
Unittestcpp

spaulaus:
Started creating unittests for the software. I am currently using the UnitTest++ framework. This may need to be changed later if we decided on a different framework.


spaulaus:
It seems that rebasing while there is an open pull request does funny things...Closing this and will open a new one later.

PR #133, opened 2016-11-29
Removing list of known detectors.

spaulaus:
This concept is antiquated, and we do not need the restriction since we have the experiment specific processors that can link into any of the known types. We still need to ensure that people are calling things by the right name for the subprocessors, but we'll have to think of a better way.


ksmith0 (approved):
Looks good. A reasonable approach for simple detectors.

PR #134, opened 2016-11-29
Updating gitignore to exclude build directories

spaulaus:


yongchix (approved):
minor changes to ignorefile

ksmith0:
I am not a fan of putting words sandwiched between wildcards. With this ignore file we are going to ignore any use the word 'build' in any file even if it is a file we may want to add in the future.

spaulaus:
We currently have an issue with people adding build directories to the base directory that are accidentally getting tracked by git. Do you have another suggestion to ignore build directories?

ksmith0:
I'm not sure what you mean? The `build/` ignores the build directory in the base level of the repo. If people make this folder at a lower level you may have problems.

I also do not have an issue with git informing you that you have put a something in a trackable location. The user does not need to commit it.

Maybe `*/build/` is a better choice?

spaulaus:
We are having issues with directories named things like `cmake-build-debug`. I agree that the user does not need to commit it, but we have already had instances in other repositories of people committing these directories by accident due to inattentive `git add`s and `git commit -a`s

ksmith0:
What goes in there? An alternate suggestion is to add `build/` again and then add an additional statement for these folders. Maybe `*-build-debug/` or `*-debug/` or `*-build-*/`.

These are the questions I think we need to ask for adding items to gitignore:
- Is this item unique to the current development work and will not be committed?
- Is the wild card statement too broad?
- Will someone pulling the repo expect that file to be ignored?

Are those reasonable items to consider? Any other suggestions? If multiple people will be creating folders like that and expect them to be ignored I would say to go ahead with your change.

spaulaus:
That's a question I'm not sure about. I think that your last suggestion `*-build-*/`, should encompass most of those types of scenarios? I agree that the `*build*` is too broad. However, we have to try and make sure that people do not unintentionally pollute the repository with things like this. Especially when those of us with knowledge of these types of things move to other positions.

ksmith0 (approved):

PR #135, opened 2016-11-30
Improved efficiency of Gsl2Fitter

tking53:


ksmith0 (dismissed):
Is there a reason you removed the virtual from the GslFitter methods?

There is no need to add `globals_` at the class scope. It is only used in `FittingAnalyzer::Analyze` and should be scoped there. This reduces the dependencies of the header.

Scan/utkscan/analyzers/source/Gsl2Fitter.cpp
```diff
@@ -89,21 +83,20 @@ void GslFitter::PerformFit(const std::vector<double> &data,
     }
 
     gsl_multifit_fdfsolver_wset (s, &f, &x.vector, &w.vector);
-    gsl_multifit_fdfsolver_driver(s, 1000, xtol, gtol, ftol, &info);
+    gsl_multifit_fdfsolver_driver(s, 100, xtol, gtol, ftol, &info);
```
ksmith0: What does this hard-coded number do? Why did it change from 1000 to 100?

spaulaus:
That number is the maximum number of iterations, the major change is reducing the tolerance from 10^-10 to 10^-4.

Scan/utkscan/analyzers/source/Gsl2Fitter.cpp
spaulaus: That number is the maximum number of iterations, the major change is reducing the tolerance from 10^-10 to 10^-4.

tking53:
i get this error when i try to compile on Quack (gsl 1.13) 

```
git-testing/paass/Scan/utkscan/analyzers/source/Gsl1Fitter.cpp: In member function ‘virtual void GslFitter::PerformFit(const std::vector<double, std::allocator<double> >&, const std::pair<double, double>&, const bool&, const double&, const double&)’:
git-testing/paass/Scan/utkscan/analyzers/source/Gsl1Fitter.cpp:108: error: ‘isFastSipm_’ was not declared in this scope
```

spaulaus:
@ksmith0 The requested changes (and more) have been made and tested.

spaulaus (approved):
The issues originally raised have been addressed. The magic number is now hard coded, but no longer magic.

PR #136, opened 2016-11-30
Removed pixie16app_defs.h from the Scan side of the project

spaulaus:
It was used for exactly two values, which are repeated and hard coded in the Unpacker class. They are not used at all in any of the utkscan related code, and are unnecessary.


ksmith0 (approved):
Looks reasonable and compiled with no issue on r230.

PR #137, opened 2016-11-30
Merging Dev into Master

spaulaus:
It's high time we actually do this. Most of the stuff in dev have gone through at least one experiment at this point. I would say that things are pretty good now. We will need to clean up the conflict.


ksmith0:
I'd like to merge #126, #127 and #128 to dev before we merge to master.

spaulaus:
That's no problem. We can shelve this one for now. If also like to get the unit tests into master as well.

PR #140, opened 2016-12-02
Unittestcpp

spaulaus:
Adds functionality for unit testing using the UnitTest++ framework.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,14 +90,6 @@ find_package (PLX)
 #Find the Pixie Firmware
 find_package (PXI)
 
-#Find the UnitTest++ Package. This package can be obtained from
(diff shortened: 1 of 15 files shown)


spaulaus:
I did a force push of the branch after a rebase on to dev to remove the merge conflict. This will prevent us from having duplicate commits in the history.

tking53:
Built and run on my laptop (Ubuntu16.10) with UnitTest++ installed to `/opt/UnitTest++`. i was also able to built and run with it (unittest++) installed under my home after adjusting cmake (passing correct path) to find the UnitTest++ lib and include directories.

sztaylor89 (approved):
Seems to be working.  Both tests fully paassed.

Issue #139, opened 2016-12-02
As a user I would like to use the colon ranges for module commands.


spaulaus:
The colon ranges ( `pwrite 0:4 4:13 TRACE_LENGTH 200` ) should be implemented with the module specific parameters as well. 

1. Colon range does not work with adjust_offsets, pmread/write, toggle, get_traces
2. csr_test does not accept hex input.

ksmith0:
Added functionality to the follwing commands int he corresponding commit:
* pmread / pmwrite: a7043a63d172b7551fb6c6bf3d2c8d8cf1ec294e
* adjust_offsets: b71f43a41792ee72f8a87da29f9fca37037ca7da
* toggle: dbe629aa144d2f37585b9607af8d7c4495b47471
* csr_test: ff14dd232ec1003faed2e88bf75000aeab67da22

`get_traces` is currently designed to only take a single channel as an argument. I would suggest we open a new issue for this as it is not as simple as the work done for the other commands.

PR #149, opened 2016-12-09
Mod / Chan Range and Hex Input

ksmith0:
Updated pmread / pmwrite, adjust_offests, toggle, and csr_test to read ranges of modules or channels and hexadecimal input.

This addresses and resolves issue #139

### DIFF
--- a/Poll/source/poll2_core.cpp
+++ b/Poll/source/poll2_core.cpp
@@ -837,7 +837,7 @@ void Poll::get_traces(int mod_, int chan_, int thresh_/*=0*/){
 /// If the attmept is unsuccesful the mehtod returns false.
 bool Poll::SplitParameterArgs(const std::string &arg, int &start, int &stop) {
 	//If a character is found that is nonnumeric or is not the delimeter we stop.
-	if (arg.find_first_not_of("0123456789:") != std::string::npos) return false;
+	if (arg.find_first_not_of("-0123456789:") != std::string::npos) return false;
 
 	size_t delimeterPos = arg.find(':');
 	try {	
@@ -845,6 +845,7 @@ bool Poll::SplitParameterArgs(const std::string &arg, int &start, int &stop) {
 		//If the delimeter was found we can seperate the stop otherwise set start = stop.
 		if (delimeterPos != std::string::npos) {
 			stop = std::stoi(arg.substr(delimeterPos + 1));
+			if (start < 0 || stop < 0 || start > stop) return false;
 		}
 		else stop = start;
 	}
--- a/Poll/source/poll2_core.cpp
+++ b/Poll/source/poll2_core.cpp
@@ -1308,8 +1308,24 @@ void Poll::CommandControl(){
 		else if(cmd == "csr_test"){ // Run CSRAtest method
 			BitFlipper flipper;
 			if(p_args >= 1){ 
-				if(!IsNumeric(arguments.at(0), sys_message_head, "Invalid CSRA value specification")) continue;
-				flipper.CSRAtest((unsigned int)atoi(arguments.at(0).c_str())); 
+				//Check that there are no characters in the string unless it is hex.
+				std::string &valueStr = arguments.at(0);
+				if (valueStr.find_last_not_of("0123456789") != std::string::npos &&
+						!((valueStr.find("0x") == 0 || valueStr.find("0X") == 0) &&
+							valueStr.find_first_not_of("0123456789abcdefABCDEF", 2) == std::string::npos) ) {
+					std::cout << "ERROR: Invalid parameter value: '" << valueStr << "'\n";
+					continue;
+				}
+				unsigned int value;
+				//Use stod to add hex capability. The decimal and negative values are
+				// caught above and rejected.
+				try { value = (unsigned int) std::stod(valueStr); }
+				catch (const std::invalid_argument &ia) {
+					std::cout << "ERROR: Invalid parameter value: '" << valueStr << "'\n";
+					continue;
+				}
+
+				flipper.CSRAtest(value);
 			}
 			else{
 				std::cout << sys_message_head << "Invalid number of parameters to csr_test\n";
--- a/Poll/source/poll2_core.cpp
+++ b/Poll/source/poll2_core.cpp
@@ -1204,11 +1204,16 @@ void Poll::CommandControl(){
 			}
 
 			if(p_args >= 1){
-				if(!IsNumeric(arguments.at(0), sys_message_head, "Invalid module specification")) continue;
-				int mod = atoi(arguments.at(0).c_str());
-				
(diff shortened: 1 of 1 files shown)


spaulaus:
When building I get the following error: 

>  CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
   Could NOT find PXI (missing: PXI_LIBRARY_DIR)
 Call Stack (most recent call first):
   /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:315 (_FPHSA_FAILURE_MESSAGE)
   cmake/modules/FindPXI.cmake:33 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
   CMakeLists.txt:124 (find_package)

I have the following line exported in my .bashrc : `export PLX_SDK_DIR=/opt/plx/current/PlxSdk`.

The PLX software is installed at `/opt/plx/current`. The current symlink points to `/opt/plx/7.1.1/PlxSdk`. (it previously pointed to `/opt/plx/7.1.1`). 

I'm working on takk. @ksmith0 : Do you see anything immediately wrong with my setup?

ksmith0:
I think you've answered your question about the PLX directory, but your error message is about the XIA software not the PLX. `PLX_SDK_DIR` on takk should be set to either `/opt/plx/current` or `/opt/plx/7.1.1/PlxSdk`.

spaulaus:
You are correct...Those two are too similar, I get them confused all the time.

ksmith0:
Perhaps we should provide some more useful error messages.

spaulaus:
I think that would be helpful. 

Ah! The issue is that the firmware folder I have "current" pointing to does not have the software directory. I think that we should make a new private repository for the API, that way we can download and update as necessary. Maybe make two separate searches : one for API, one for firmware.

ksmith0:
That is basically what is happening now. The current symbolic link is just a guess in the search. The user can manually specify the folder's location. The following code from FindPXI.cmake searches for the library files and determines the include directories based on its location. 

```cmake
#Find the library path by looking for the library.
find_path(PXI_LIBRARY_DIR
	NAMES libPixie16App.a libPixie16Sys.a 
	HINTS ${PXI_ROOT_DIR}
	PATHS /opt/xia/current /opt/xia/software
	PATH_SUFFIXES software
	DOC "Path to pixie library.")

if (PXI_FOUND)
	set (PXI_INCLUDE_DIR ${PXI_LIBRARY_DIR}/inc ${PXI_LIBRARY_DIR}/sys ${PXI_LIBRARY_DIR}/app)
	set(PXI_LIBRARIES -lPixie16App -lPixie16Sys)
endif()
```

In addition, the library directory is used to infer the location of the firmware if not specified via `PXI_FIRMWARE_DIR` as follows.

```cmake
get_filename_component(PXI_LIBRARY_DIR "${PXI_LIBRARY_DIR}" REALPATH)

if(NOT PXI_FIRMWARE_DIR)
	get_filename_component(PXI_FIRMWARE_DIR "${PXI_LIBRARY_DIR}/../.." REALPATH)
endif(NOT PXI_FIRMWARE_DIR)
set(PXI_FIRMWARE_DIR ${PXI_FIRMWARE_DIR} CACHE PATH "Path to folder containing XIA firmware.")
```

Finally `PXI_CONFIG` makes use of the `PXI_FIRMWARE_DIR` to recursively search for directories with dsp and firmware folders.
```cmake
#Look in the root directory for the XIA library
	if(NOT EXISTS ${PXI_FIRMWARE_DIR})
		message(WARNING "Configuration Error - Invalid Pixie firmware directory: ${PXI_FIRMWARE_DIR}")
		return()
	endif(NOT EXISTS ${PXI_FIRMWARE_DIR})
	subdirlist(PXI_FIRMWARE_DIRS ${PXI_FIRMWARE_DIR})

	#remove directories without subdirectories firmware and dsp.
	foreach(FIRMWARE_DIR ${PXI_FIRMWARE_DIRS})
		if (NOT (EXISTS ${FIRMWARE_DIR}/firmware OR EXISTS ${FIRMWARE_FIR}/dsp))
			list(REMOVE_ITEM PXI_FIRMWARE_DIRS ${FIRMWARE_DIR})
		endif (NOT (EXISTS ${FIRMWARE_DIR}/firmware OR EXISTS ${FIRMWARE_FIR}/dsp))
	endforeach(FIRMWARE_DIR ${PXI_FIRMWARE_DIRS})
```

spaulaus:
I see. We should add this as an additional in the installation process. Maybe reconfigure the `/opt/xia` directory to have two subfolders firmware and software.

spaulaus:
@ksmith0 Why do I get OK message for Good channel and Hi/Lo gain? 
When performing a csr_test with the hex input I found that it was giving me a `...[OK]` on the Good channel and High/Lo gain lines for the test output. I cannot replicate this error at this time.

spaulaus:
I'm having some weird issues with the terminal right now. The backspace key just advances me forward, when trying to edit a line the cursor sometimes jumps to the line above. This has been an issue with SSH'd sessions in the past. Everything works, I will confirm that this issue is not present on the actual computer tomorrow.

ksmith0:
Could you run the terminal debugger and see what character codes are being sent when you type backspace?

spaulaus:
What's the command for that?

ksmith0:
Build with `BUILD_TESTS=ON` then run `CTerminalTest` and do something like the following:

```
This is test terminal.
Test>   
help    quit    debug   tab
Test> debug
Debug mode is on.
TERM: Curs (6, 43) Key: 104 h
TERM: Curs (7, 43) Key: 101 e
TERM: Curs (8, 43) Key: 353 a
TERM: Curs (9, 43) Key: 9
TERM: Tab complete 0<->3  'hea  '
TEST: cmd='hea  ' arg=''
Test> hea
```

ksmith0:
>Why do I get OK message for Good channel and Hi/Lo gain?

I have no idea what you mean.

ksmith0:
Your backspace issue may be related to #70.

spaulaus:
You are correct, I couldn't find that issue number when I searched before. 

The terminal debugger provides 
For backspace key : `TERM: Curs (6-6, 22) Key: 8`
For Delete key : `TERM: Curs (7-6, 22) Key: 330 J`

I amended my comment for the OK message.

spaulaus (approved):
Since the backspace error is a known bug. We can ignore it.  I will add my comment to the appropriate Issue. The issue with the misplaced OK sign is not repeatable and could have been due to the small terminal size I had.

ksmith0:
You can check the OK issue by looking in your log file and seeing if the display is correct. I have a strong suspicion that it may have been related to your terminal issues.

spaulaus:
We should be clear that my terminal works fine....

PR #150, opened 2016-12-12
Updated terminal to handle ASCII codes for BS/DEL.

ksmith0:
ncurses.h does not have the ASCII codes fro backspace and delete. Added
a check for those in addition to the codes in ncurses.h to handle these
keys.

This resolves issue #70.


spaulaus (changes requested):
This did not fix the issue. Please see the video at the following address: 
https://drive.google.com/open?id=0B_XQp_kYuMwQbkNHTVFleUx1NVU

spaulaus:
I will reopen this in a bit.

PR #151, opened 2016-12-12
Updated CTerminalTest to display TERM.

ksmith0:
The CTerminalTest program now displays the TERM environment variable when starting.
Also, changed to static library linkage.


spaulaus (approved):
Works as expected

PR #152, opened 2016-12-14
Gitignore update

tking53:
I updated the gitignore to exclude the doxywarning.txt file created when generating the Doxygen files. I also updated the doc/doxyfile to exclude the install directory when generating the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@ latex/
 tmp/
 doc/classdiagramRawEvent.svg
 ValgrindOut.xml
+doc/
+doxywarning.txt
 
 #Compressed Files
 *.bz2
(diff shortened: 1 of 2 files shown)


spaulaus:
Will this also ignore any other variations on the install directory? For example, `install-gitignoreUpdate`, `test-install`, etc.

tking53:
It does not, would you like it too? I was trying to avoid being to general.

spaulaus:
I think in this case we have to be general. We never want the install directory to be found as part of the documentation. The contents of this directory are all contained within the rest of the source trees, which would always result in duplication.

spaulaus (approved):
The Doxygen build has been updated along with the gitignore.

PR #153, opened 2016-12-19
Abstracted ROOT canvas from scope to RootScanner.

ksmith0:
The ROOT canvas and zoom functionality has been moved to RootScanner
class and scope now inherits from this class.


spaulaus (approved):
Worked without issues.

PR #154, opened 2016-12-21
Newcfd

spaulaus:
Adding the new CFD algorithm developed by @cthornsb and @ksmith0. I also added in the TraditionalCfd. I moved a lot of classes to the new Resources folder so that they could be used by any of the programs in the Scan directory. We may want to move some of these even further up. This rewrite has increased the efficiency of the trace analysis which can be seen in an overall decrease in the amount of time spent in the WaveformAnalyzer.


spaulaus:
I appreciate it if we could get this tested and approved quickly. This starts some of the reorganization that is necessary to update other things like the data classes, globals, trace analysis.

ksmith0 (approved):
Looks good, the new cfd appears to provide reasonable results. Unit tests appear to check for what would be expected.

PR #155, opened 2017-01-03
Updated terminal to handle ASCII codes for BS/DEL.

ksmith0:
ncurses.h does not have the ASCII codes fro backspace and delete. Added
a check for those in addition to the codes in ncurses.h to handle these
keys.

This resolves issue #70.

Former-commit-id: 1c13ce2ef7feb7aac6cdd480559047ff844e0a08


spaulaus (approved):
Works for most of us.

PR #156, opened 2017-01-04
Decoding update

spaulaus:


ksmith0 (approved):
Looks reasonable, has passed all unit tests. Has been used to unpack the data associated with the newcfd.

PR #158, opened 2017-01-05
Added GetOutputName to globals

tking53:
This replaces the old GetArguments way of using the .his name as the name of the created root file. I have been using this on my branches for a while now with no issues.


spaulaus (changes requested):
Some of the spacing needs to be cleaned up. I don't see any functional issues with this, since I've implemented the same thing  in one of my personal branches.

spaulaus (approved):

PR #159, opened 2017-01-06
Merge Release candidate into master

spaulaus:
The RC branch was forked at commit ID 773c297a4c720ef49e49bccf2f5c3dd0dea63e6a. The e14060 branch was also forked from this ID. The experiment ran without any difficulties from the acquisition or analysis side. I do not see any reason that we could not update the master branch to this point.

### DIFF
--- a/Scan/utkscan/core/include/GetArguments.hpp
+++ b/Scan/utkscan/core/include/GetArguments.hpp
@@ -1,0 +1,43 @@
+/** \file GetArguments.hpp
+ * \brief A quick and dirty way to get the arguments pass to scan through
+ * the respective Fortran calls
+ *
+ * \author David Miller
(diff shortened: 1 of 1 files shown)


ksmith0:
Should I be worried about the test failures?

spaulaus:
Ah, that's shippable. Ignore those failures. I need to remove the integration.

ksmith0:
I will approve this, but I'm not sure what the correct process is here. Do we need to have a group of people testing this? Should we require more than one review?

ksmith0 (approved):
This should be a reasonable release.

spaulaus:
I think the process would be more straight forward since the RC should only be pushed from dev after the features have undergone testing for a while.

PR #165, opened 2017-01-12
Hribf fixes

cthornsb:
Contains various file header fixes for ldf and pld file formats.


spaulaus:
Didn't rebase before pull request. Will need to start a new one.

PR #167, opened 2017-01-13
Removing throw for Bad Event Length error

spaulaus:
Due to issues with data corruption for data taken using PixieSuite2 this throw caused many data files to be unreadable. We have recovered previous functionality and updated the error message accordingly.


tking53 (approved):
code compiles and runs. no longer crashes on a bad buffer

spaulaus:
I should note that the code never crashed on a bad buffer. It did exactly what it I told it to do, which was throw an exception.

PR #170, opened 2017-01-13
Now properly handling the sign of the x^2 coefficient

spaulaus:
Fixes an issue identified by @tking53 where the analysis would fail since we didn't choose the right root.


ksmith0 (approved):
Matches what we discussed.

tking53:
runs, seems to give same 3245 as fitting.

Issue #168, opened 2017-01-13
TreeCorrelator getting bad place


spaulaus:
The TreeCorrelator is attempting to get a bad place in the DetectorDriver. We are checking for `__-1` but getting `__4294967295`. This is a signed vs unsigned int issue, that I am currently having trouble tracking down. The error is occurring at line 324 in detector driver.

ksmith0:
From dev branch [ paass/Scan/utkscan/core/source/DetectorDriver.cpp](https://github.com/pixie16/paass/blob/dev/Scan/utkscan/core/source/DetectorDriver.cpp)
```c++
string place = (*it)->GetChanID().GetPlaceName();
if (place == "__-1")
  continue;
```
Where `(*it)` is a `ChanEvent*`.

`ChanEvent::GetChanID ()` returns an `Identifier&`

`Identifier::GetPlaceName()` calls `Identifier::GetLocation()` as follows:
```c++
/** \return The name of the place associated with the channel */
std::string GetPlaceName() const {
    std::stringstream ss;
    ss << GetType() << "_" << GetSubtype() << "_" << GetLocation();
    return ss.str();
}
```

`Identifer::GetLcoation()` returns `Identifier::location` which is set to -1 in `Identifier::Zero()`

I suspect the following, but have not yet made the connection:
`Identifier::GetTag()` returns an `int` equal to `std::numeric_limits<int>::max()` if the tag is not found. This value is equal to `INT_MAX` which is not -1. You are reporting that you are getting `U LONG_MAX` which should be the max value for an `unsigned long int`. http://www.cplusplus.com/reference/climits/

spaulaus:
Nailed it. Thanks @ksmith0

ksmith0:
I still don't see how your commit resolved the issue. You changed the default value from -1 to 9999 (did not update the if statement in DetectorDriver) and removed the Getter calls in GetPlaceName, which should have been returning the values as you changed them to. Should this behavior not just be the same?

spaulaus:
You are correct. I forgot to stage the DetectorDriver.

ksmith0:
I understand now. The `location` was an `unsigned` thus setting it to -1 causes it to set to the maximum value. Changing to 9999 ensures a value that can be checked against. This does not permit a user to a 10,000 channel system and it may have been better / easier to simply change the if statement to check against the largest possible value of `unsigned int`. The issue was resolved with your commit and has been merged.

spaulaus:
[...]

PR #171, opened 2017-01-13
Fixing issue #168

spaulaus:
I did some formatting as well. Important changes on line Identifier.hpp:142 and Identifier.cpp:20-21.

### DIFF
--- a/Scan/utkscan/core/include/Identifier.hpp
+++ b/Scan/utkscan/core/include/Identifier.hpp
@@ -37,43 +37,43 @@ public:
      * \param [in] loc : the location to set */
     Identifier(const std::string &atype, const std::string &subType,
                const unsigned int &loc) {
-        type = atype;
-        subtype = subType;
-        location = loc;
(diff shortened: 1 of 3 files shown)


ksmith0:
I don't understand how this resolved the issue, see comment in #168.

spaulaus:
I neglected to stage DetectorDriver.

ksmith0:
All you've done is change the -1 to 9999. You've haven't addressed where the large number is coming from.

spaulaus:
This fixes the bug that was causing the crash. In order to fix the larger issue with the TreeCorrelator and Places requires significantly more work. We can open up a separate issue to refactor these classes, but I do not see the point right now.

tking53:
The patch does allow me to scan one of Steven's ldfs.

sztaylor89 (approved):
Fixes exceptions when channels aren't defined in the map.  The 9999 default value may cause issues for systems with 10,000 channels.

PR #172, opened 2017-01-13
If the USE_ROOT Flag was turned off, then we couldn't compile ScanLib

spaulaus:
Fixes compilation error if the USE_ROOT Flag was turned off.


tking53 (approved):
Fixes the issue.

PR #173, opened 2017-01-13
Experiment processor update

spaulaus:
Adding more of the experiment specific processors to tracking. This means the users are less likely to have to edit DetectorDriver or the CMakeLists.txt to use "standard" processors.


tking53 (approved):
code compiled without issue.

Issue #160, opened 2017-01-06
scope Segfaults when scanning 


tking53:
here is the gdb output
![scope-issue](https://cloud.githubusercontent.com/assets/18449805/21722719/d5ba8a64-d3fb-11e6-9cdb-5b2f315ff0e2.png)

spaulaus:
@tking53 Was this ever addressed?

tking53:
Nope not yet.

spaulaus:
It's fixed now.

PR #174, opened 2017-01-13
Fix to prevent scope from segfaulting.

spaulaus:
Fixes an unreported issue that was causing scope to segfault when running.


tking53:
I believe this fixes #160.

tking53 (approved):
Code compiles and runs. I can run scope with Ornl2016 Rb data which was crashing scope earlier this week.

PR #175, opened 2017-01-13
Fixing bug in the Waveform and Fitting Analyzers

spaulaus:
This bug is probably the cause for the updside down histograms.


tking53 (approved):
Compiled and run without issue.  It does indeed fix the upside down histograms.

PR #176, opened 2017-01-17
Fixing error on the extrapolated maximum

spaulaus:
The alteration here broke the performance of the CFD quite a bit. This fixes the original functionality.


ksmith0:
What was the previous commit that broke this functionality?

spaulaus:
It was this commit : 037ad6a4324db6e3f3b74a760405714770f832a9. 

Calculating the extrapolated maximum from the baseline subtracted trace is not the same as calculating the extrapolated maximum from the raw trace then subtracting the baseline. The issues were not readily apparent in the test data. But appeared after further testing.

ksmith0 (approved):

PR #188, opened 2017-01-23
Fixes #184 - Forces frequency to be a required flag for scan codes 

### LINKED ISSUE #184: Scope does not require frequency, but crashes upon running.
When scope is evoked as 
```
scope -f R34688 -i /data/HAGRiD/run_001.ldf
```
It immediately exits with the following message:
```
Scope $ terminate called after throwing an instance of 'std::invalid_argument'
[READ] 1302 words (0%), GOOD = 1, LOST = 0                                      what():  XiaListModeDataMask::GetCfdFractionalTimeMask : Could not obtain a mask for firmware code 4 and frequency 0. Check your settings.
                                                                    Aborted (core dumped)
```

I believe the proper treatment should be that scope complains about the frequency not being defined and not start.

spaulaus:
Fixes this issue.

### DIFF
--- a/Scan/ScanLib/source/ScanInterface.cpp
+++ b/Scan/ScanLib/source/ScanInterface.cpp
@@ -12,6 +12,7 @@
  */
 #include <iostream>
 #include <sstream>
+#include <stdexcept>
 #include <thread>
 
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


ksmith0 (approved):
Works as expected.

Issue #185, opened 2017-01-23
pulser.xml does not work with master utkscan


ksmith0:
The file `share/utkscan/cfgs/pulser.xml` does not run with the master version of utkscan. The error message is:
```
   * Loading PulserProcessor                                     
                                                                  [FAIL]
Exception caught at DetectorDriver::DetectorDriver
	DetectorDriver: unknown processor typePulserProcessor
```

I believe the `PulserProcessor` has been removed and this file was not updated / removed.

spaulaus:
That is correct. I am going to label this as a wontfix and close, since this same complaint could be applied to 90% of the configurations.

ksmith0:
Can we then remove these config files? They confuse those who are not intimately familiar with the history of the scan code.

spaulaus:
No, *should* be updated, but this is a rather massive undertaking. Especially since the format of these files is getting ready to change. Removing that one file means we have to remove all the others on the same argument.

spaulaus:
I would suggest that the example.xml be updated, and make sure that all users use that as a base.

PR #189, opened 2017-01-23
Partial fix for #185 - Updating example.xml so that it contains the latest information.

spaulaus:
Partially addresses issue #185


spaulaus:
I'm going to close this one, since it's really a larger issue.

PR #194, opened 2017-01-25
Fixes #122 - GeneralException from TemplateExpProcessor with no TreeCorrelator node

### LINKED ISSUE #122: GeneralException from TemplateExpProcessor with no TreeCorrelator node.
When omitting the TreeCorrelator node from a config file as below or removing it completly.

``` xml
    <TreeCorrelator name="root" verbose="False">
<!--
        <Place type="PlaceOR" name="Beta" fifo="10">
            <Place type="PlaceThreshold" name="template_start_0"
                   low_limit="10.0" high_limit="16382" fifo="5"
                   replace="true"/>
        </Place>
-->
    </TreeCorrelator>
```

The utkscan throws a general exception due to the following piece of code in `TemplateExpProcessor::Process`

``` c++
bool isTapeMoving = TreeCorrelator::get()->place("TapeMove")->status();
```

which calls the following

``` c++
Place* TreeCorrelator::place(std::string name) {
    map<string, Place*>::iterator element = places_.find(name);
    if (element == places_.end()) {
        stringstream ss;
        ss << "TreeCorrelator: place " << name
           << " doesn't exist " << endl;
        throw TreeCorrelatorException(ss.str());
    }
    return element->second;
}
```

The `throw` is called since the place "TapeMove" does not exist. This is then caught by 

``` c++
    } catch (GeneralException &e) {
        /// Any exception in activation of basic places, PreProcess and Process
        /// will be intercepted here
        cout << "Exception caught at DetectorDriver::ProcessEvent" << endl;
        cout << "\t" << e.what() << endl;
        exit(EXIT_FAILURE);
   }
```

which exits the scan program abruptly leaving the user without an understanding of the reason due to a loss of the message since the CTerminal is improperly destroyed. 

In addition, if the cmake flag `USE_ROOT=ON` is set the ROOT objects are not properly cleaned up and a segmentation fault occurs.

spaulaus:
I have replaced the exits that occur during the exceptions with throw. This will throw the errorup until it reaches main, where we handle exiting cleanly by returning. This fixes the issue where the terminal would get corrupted.

### DIFF
--- a/Scan/utkscan/CMakeLists.txt
+++ b/Scan/utkscan/CMakeLists.txt
@@ -70,7 +70,8 @@ else(USE_HRIBF)
 endif(NOT USE_HRIBF)
 
 #Add libraries to be linked with utkscan
-target_link_libraries(${SCAN_NAME} ${LIBS} ScanStatic ResourceStatic)
+target_link_libraries(${SCAN_NAME} ${LIBS} ScanStatic ResourceStatic
+        PixieCoreStatic)
 
 #If we have GSL installed link
 if(USE_GSL)
--- a/Scan/utkscan/core/source/DetectorDriver.cpp
+++ b/Scan/utkscan/core/source/DetectorDriver.cpp
@@ -20,6 +20,7 @@
 #include "DammPlotIds.hpp"
 #include "DetectorDriver.hpp"
 #include "DetectorLibrary.hpp"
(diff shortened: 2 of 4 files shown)


ksmith0 (approved):
A review of the code looks like this is simply a change from exit to throw.

PR #195, opened 2017-01-25
Fixes #121 - Empty TreeCorrelator node causes problems with GeProcessor

### LINKED ISSUE #121: Empty TreeCorrelator node causes problems with GeProcessor.
Empty TreeCorrelator node with GeProcessor and no Ge detectors in map causes utkscan to hang at startup. On my personal laptop the memory gets consumed, on kqxhc it continues to run for 2+ minutes.
This is a bad config file: [config.xml.txt](https://github.com/pixie16/paass/files/520102/config.xml.txt)

spaulaus:
This was fixed by wrapping the calls to the TreeCorrelator in try/catches. Once the pull request for #122 is merged this will no longer cause a premature exit.

### DIFF
--- a/Scan/utkscan/processors/source/GeProcessor.cpp
+++ b/Scan/utkscan/processors/source/GeProcessor.cpp
@@ -18,6 +18,7 @@
 
 #include "DammPlotIds.hpp"
 #include "DetectorLibrary.hpp"
+#include "Display.h"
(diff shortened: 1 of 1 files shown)


ksmith0 (approved):
Moved to try / catch constructions. This should reduce number of fatal crashes.

PR #197, opened 2017-01-25
Tells the user where to go to find more information about the firmware.

spaulaus:
Explicitly points them to the wiki when `-h` is called.


ksmith0:
This is an improvement. I'd like to see something like `utkscan help firmware` and you get a long message about the firmware choices. The internet is not always available.

spaulaus:
You want us to actually parse those two arguments?

ksmith0:
Yes. I think having something along these lines is extremely useful. We could implement it as an option passed to a flag `utkscan -h firmware`

spaulaus:
The `-h` handling optional arguments might be tricky. I will look into it.

spaulaus:
It looks like the getopts isn't wanting to parse the optional argument properly. I think that if we want to implement something like this, then we're going to have to implement our own class to parse the command line options.

spaulaus:
@ksmith0 , 

I have updated the help output to tell the users that they can locally find the information in HelperEnumerations.hpp. I didn't specify relative path since it would take up too much space. I think that this is as best I can do for now. I tried for an hour and a half to get the long output option working, but ran into a bunch of buggy issues with the getopt_long.

LTS is to write our own class, which we've discussed in the past.

ksmith0:
I appreciate you looking into trying to handle the help arguments. I propose we move this upgrade to an issue and I'll examine this pull request.

spaulaus:
I had to adjust for changes in the directory structure. I wasn't sure how to handle this in the rebase and ended up with an additional commit to move the old file into the new structure.

ksmith0 (approved):
Works when tested with scope.

PR #200, opened 2017-01-30
Moved INSTALL_PREFIX definition.

ksmith0:
This allows faster recompile times as less objects are set as dependent
on this compiler defintion.


spaulaus (approved):
Compiles fine, doesn't seem to break anything.

Issue #72, opened 2016-09-07
Modify set2ascii to provide correct tau values.


ksmith0:
The TAU value is stored in the set file as an IEEE standard. We are incorrectly treating it in set2ascii.

> Hi Mitch,
> 
> Stan is correct and you also figured it out already: the preamp tau has the IEEE standard format. The reason for that is we need to store it in the format which allows decimal fractions, e.g. 46.25 us. The only other number that has the same IEEE format is the baseline values in the list mode data, since the baseline can have decimal factions too.
> 
> Regards,
> Hui

spaulaus:
This also occurs in the data decoding of the Baseline value.

PR #202, opened 2017-01-31
Directory overhaul to enhance clarity for users

spaulaus:
I  overhauled the directory structure to make things clearer for the users. We now group the Acquisition (poll, MCA, etc.) into a unified folder as it has been with Scan (renamed to Analysis). This necessitated a number of changes to the CMakeLists.txt files for these directories.  

I  renamed some of the static and shared libraries so that they are a little bit clearer. For example, PixieCoreStatic is now PaassCoreStatic. I think it makes things a little clearer since we've got other things named "Core" floating around. 

I created the Resource and ThirdParty directories. These directories house software that are useful to many different pieces of the software. Previously, the software contained within these directories was contained solely in the Scan (now Analysis) directory. This will allow fixes for things like #72 to be more useful, since the code necessary to fix this can be applied to unpacking the data as well. 

The ThirdParty software compiled with the code is only pugixml. However, if we move to the Google Testing framework then we would be able to add this software here. This also makes it possible to include some dedicated minimization software that doesn't rely on GSL or ROOT. 

P.S. We need a label for Affects literally everything.


CMakeLists.txt
```diff
@@ -65,17 +69,16 @@ option(USE_ROOT "Use ROOT" ON)
...
-
 #The MCA will write DAMM histograms as output
 if(USE_DAMM)
     add_definitions("-D USE_DAMM")
 endif()
 
 #------------------------------------------------------------------------------
 
-#Find packages needed for poll2
+#Find packages needed for the
```
ksmith0: for the ... ?

CMakeLists.txt
```diff
@@ -112,8 +115,7 @@ if (BUILD_SUITE)
 	else (PLX_FOUND OR PXI_FOUND)
 		#Find the PLX Library
 	   find_package(PLX REQUIRED)
-   	link_directories(${PLX_LIBRARY_DIR})
-	   #add_definitions("-DPLX_LINUX -DPCI_CODE -DPLX_LITTLE_ENDIAN")
```
ksmith0, line 116: This line was here as carried over from PixieSuite1. Do we know what it did and that it was safe to remove them?

CMakeLists.txt
```diff
@@ -141,38 +143,30 @@ if (USE_ROOT)
...
-	if (BUILD_SETUP) 
-		include_directories(Setup/include)
-		add_subdirectory(Setup)
-	endif()
-
-	#Build poll
-	add_subdirectory(Poll)
+#Build Acquisition software, we don't have this broken into as fine of a
+#granularity as the Analysis software, so we just wrap the whole thing in an if.
+if (BUILD_SUITE)
```
ksmith0: Should we change BUILD_SUITE to BUILD_ACQ?

CMakeLists.txt
spaulaus: It was commented out. I thought I just removed the comment. I will add back in the link_directory line.

ThirdParty/source/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(ThirdPartySources pugixml.cpp)
+
+add_library(ThirdPartyObjects OBJECT ${ThirdPartySources})
+add_library(ThirdPartyStatic STATIC $<TARGET_OBJECTS:ThirdPartyObjects>)
```
ksmith0: I'm not sure these names make sense. I doubt you will often want to link against all the third party things at once. One library for each makes more sense to me.

CMakeLists.txt
ksmith0: Yes it was commented out because we were not sure of its function. I don't want to throw it away unless we know we don't need it.

CMakeLists.txt
spaulaus: You know, I had started to do that, but changed my mind midway through. I stopped because I didn't want to figure out what/how to change things for the Scan part.  **I'll change it. It makes sense.**

ThirdParty/source/CMakeLists.txt
spaulaus: No problem. I'll make the change.

CMakeLists.txt
spaulaus: shut up? I'll fix it.

CMakeLists.txt
spaulaus: If we've gotten this far without it, then I suspect that we do not need it. I would prefer not to have blocks of code commented out.

spaulaus:
@ksmith0 : I have addressed each of your comments above. I have not added back in the commented line since I'm not convinced that we need it.

ksmith0 (approved):
Compiled with cmake 2 and 3 on r230 and takk. With and without HRIBF.

PR #203, opened 2017-02-01
Updating doxyfile to accommodate changes to directory structure.

spaulaus:


Doc/doxyfile
```diff
@@ -771,8 +771,8 @@ WARN_LOGFILE           = doxywarning.txt
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ./Core cmake doc/pages/ Interface MCA Poll \
-                         PxiDump Scan Setup
+INPUT                  = Acquisition Analysis Core Cmake Doc/pages/ \
```
ksmith0, line 774: Doc/pages?

Doc/doxyfile
```diff
@@ -812,11 +812,8 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = Scan/utkscan/core/include/pugiconfig.hpp  \
-                       Scan/utkscan/core/include/pugixml.hpp \
-                       Scan/utkscan/core/source/pugixml.cpp \
-                       Scan/utkscan/include/MersenneTwister.hpp \
-                       README.md \
+EXCLUDE                = Analysis/Utkscan/core/include/MersenneTwister.hpp \
```
ksmith0, line 815: Oooh? What's this?

spaulaus: This is the header that contains an implementation of the MersenneTwister algorithm. It was implemented to provide the random numbers that we add to the energies from pixie. Its original purpose was because C++ didn't have any useful built in random number generators. 

Since we've adopted C++ Std 11 we could remove this file, or at least move it to the base Resources folder. However, I've just not gotten around to tackling that problem.

ksmith0 (approved):
Works to produce the doxygen documentation as expected.

PR #205, opened 2017-02-01
Fixed incorrect DATA buffer size due to spill footer

cthornsb:
Previously, the ldf DATA buffer writer incorrectly
handled occurances where the spill footer could not
fit in the current buffer. This was causing certain
DATA buffers to have a number of words less than
8194 and, in turn, causing the ldf reader to get
out of position.


ksmith0 (approved):
Resolves the issue with threshold set to 0.1 on gio. Tested with parent commit and had issue and this commit resolved it.

PR #206, opened 2017-02-01
Resolves #72 - Modify set2ascii to provide correct tau values

### LINKED ISSUE #72: Modify set2ascii to provide correct tau values.

spaulaus:
We added a method to convert from IEEE Floating point standard number to human readable decimal number. Added a unit test for this function, and fixed the output in set2ascii. We did this so that we could use these number in calculations, and verification of settings.

### DIFF
--- a/Acquisition/PxiDump/source/set2root.cpp
+++ b/Acquisition/PxiDump/source/set2root.cpp
@@ -17,6 +17,7 @@
 #include "TFile.h"
 #endif
 
+#include "HelperFunctions.hpp"
 #include "set2root.hpp"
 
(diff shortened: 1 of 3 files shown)


ksmith0 (approved):
Tested on r230, works as expected.

PR #209, opened 2017-02-03
Fixes #207 - Missing <stdexcept> header include

### LINKED ISSUE #207: Missing <stdexcept> header include
PixieInterface.cpp and poll2_core.cpp use std::invalid_argument, but the header (#include <stdexcept>) is not included in these files. This caused the dev branch of paass to not install on NItemp.

spaulaus:
These missing headers caused compilation issues on some computers. Fixes #207.


ksmith0 (approved):
Compiled on r230 with gcc 4.8.5 no warnings. This should resolve issues on nitemp.

PR #211, opened 2017-02-03
Updated CMake

ksmith0:
Changed to ../install from build and amended FindPXI.cmake messages.


spaulaus (commented):
This did what it said it did. 

Tested on 
Linux Flavor Arch Linux. 
Kernel : 4.9.6-1-ARCH
gcc version:  6.3.1 20170109 (GCC)
cmake version 3.7.2

spaulaus (approved):
Still does what it says. More importantly, it removes that annoying message.

Issue #146, opened 2016-12-07
Sort auto-configured pixie.cfg file.


ksmith0:
This idea came up during merge request #128:

>4. Sorting pixie.cfg by revision (?) This may not be critical but would be nice.

This is more work as the ModuleType is determined after entering a directory and the output is written out while parsing the files inside that directory. To do as you ask one would need to store the output into a string (can you store multiple lines in a cmake string?). This string is somehow associated with the tag then you would need to sort the tags (not sure this is possible in cmake) and output the associated string.

In addition, one would want to sort by revision, then adc resolution, then sample speed.

spaulaus:
Do we really even need this? It would be nice, but as long as things are clearly documented then it shouldn't matter.

ksmith0:
It was your suggestion, but it is very difficult. I agree with you that things are clear enough. Closing as wontfix.

spaulaus:
I remember suggesting it, but have rethought it.

Issue #55, opened 2016-08-29
make config by default


ksmith0:
Duplicate of pixie16/PixieSuite2#75.

PR #213, opened 2017-02-03
Renamed FindPXI to FindXIA

ksmith0:
Also, renamed all PXI variables to XIA.

Also, attempted to resolve Issue #146, but it was too difficult for the time allotted.

### DIFF
--- a/Acquisition/Interface/source/CMakeLists.txt
+++ b/Acquisition/Interface/source/CMakeLists.txt
@@ -2,8 +2,8 @@ set(Interface_SOURCES PixieInterface.cpp Lock.cpp)
 
 add_library(PixieInterface STATIC ${Interface_SOURCES})
 
-#Order is important, PXI before PLX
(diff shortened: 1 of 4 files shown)


spaulaus (approved):
I did not find any issues here. 

Tested on 
Linux Flavor: Arch Linux
Kernel : 4.9.7-1-ARCH
cmake : 3.6

ksmith0:
@spaulaus Did you test this on a machine that would attempt to look for the XIA libraries? Did you try `make config`?

spaulaus:
Mine does look for XIA libraries. I didn't test make config. I thought we had set the config to be automatically run with install?

ksmith0:
We did not change `make config` to default. That is issue #55.

spaulaus:
Ah! I just checked and make config is fine. 
```
Scanning dependencies of target config
-- Installing: install/share/config/pixie.cfg
Built target config
```

PR #214, opened 2017-02-03
Changed CTerminal to only remove prompt on enter.

ksmith0:
When a command is entered the prompt is cleared until GetCommand is
called again. The KEY_UP and KEY_DOWN commands call clear_ and required
printing of the prompt after the clear_ as well. This permits long running commands such as `adjust_offsets` to indicate to the user that something is happening.


ksmith0:
Prior to this the prompt would reappear immediately adfter the command was entered. If a command took some time to complete it appeared as if nothing happened and the user would retype the command to be surprised by the eventual output.

spaulaus:
I think this is now too far out of date to fix properly. I'm going to close this PR and start a new one. However, i can confirm that this works as described.

spaulaus:
@ksmith0, can you update this branch to the latest HEAD of dev and resubmit the request? I can then approve it. I suggest also adding a long running command to the CTerminalTest so that we can test features like this in the future.

ksmith0:
This need to be rebased or cherry-picked before being merged.

spaulaus (changes requested):
Needs rebased and updated before merging.

Issue #76, opened 2016-09-15
Monitor column label


ksmith0:
Someone asked me about column labels for the monitor. It's not a bad idea.

PR #215, opened 2017-02-03
Added headers to the columns in monitor.

sburcher:


ksmith0 (approved):
Very helpful for those less familiar with monitor. Works as expected. Tested on r230 with up to 3 modules.

ksmith0:
This resolved #76.

Issue #106, opened 2016-09-23
HighResTimingData should be child of ChanEvent/ChannelData


spaulaus:
This references issue #2, but expands the logic to include the HighResTimingData class. It is more of an extension of the ChannelData/ChanEvent/Whatever class. I believe this also has to do with some of the problems mentioned in #87.

Issue #87, opened 2016-09-17
Issue getting Walk Corrected Times


spaulaus:
The walk corrected times seem to be different than the uncorrected times when there is no correction applied to the channel. This has only been confirmed for the HighResTimingData class.

Issue #16, opened 2016-06-23
48-bit times in RawEvent


spaulaus:
We should consider the need to have longer time stamps than 32bit. This does not have to be limited to 48 bits.

PR #218, opened 2017-02-06
Resolves Issue #106, #2, #16 and #87 - Rewriting ChannelData, ChannelEvent, etc. 

spaulaus:
This pull request addresses issues #106, #2, #16 and #87 

All of the data classes are now derived from the XiaData class either directly (Processed XiaData) or through several generations. We no longer need a bunch of wrapper functions in HighResTimingData, or any of the other data classes since we are able to access any of the data through the methods defined by the base classes. 

All of the updates to the various processors was to ensure that these data were being accessed properly and for cleanup. 

During testing it became necessary to have the  XiaDataDecoder class store not only the arrival time of the signal but also the Time sans CFD information. Offline timing needs the filter time sans CFD information.


spaulaus:
I'm going to close this until further testing of this branch is complete. I do not think that everything is working as it should be.

PR #219, opened 2017-02-06
Resolves #82 - Linking error USE_GSL=OFF

### LINKED ISSUE #82: Linking error USE_GSL=OFF
Built dev branch on takk (gcc 4.8.5) with following cmake flags

```
BUILD_UTKSCAN=ON
USE_GSL=OFF
```

gives this error:

```
Scanning dependencies of target utkscan
Linking CXX executable utkscan
core/CMakeFiles/CoreObjects.dir/DetectorDriver.cpp.o: In function `DetectorDriver::LoadProcessors(Messenger&)':
DetectorDriver.cpp:(.text+0x655f): undefined reference to `FittingAnalyzer::FittingAnalyzer(std::string const&)'
collect2: error: ld returned 1 exit status
make[2]: *** [Scan/utkscan/source/utkscan] Error 1
make[1]: *** [Scan/utkscan/source/CMakeFiles/utkscan.dir/all] Error 2
make: *** [all] Error 2
```

spaulaus:
Adding in new CMake module to allow for dependent options.
Including this module allows us to have the USE_GSL option depend on the BUILD_UTKSCAN option. We no longer have the issue where the two are decoupled and this resolves the error that we previously had.

### DIFF
--- a/Analysis/CMakeLists.txt
+++ b/Analysis/CMakeLists.txt
@@ -1,5 +1,6 @@
 option(USE_HRIBF "Use HRIBF library for scan base." OFF)
-option(USE_GSL "Use GSL for Pulse Fitting" OFF)
+CMAKE_DEPENDENT_OPTION(USE_GSL "Compile with GSL" ON "BUILD_UTKSCAN" OFF)
+mark_as_advanced(USE_GSL)
 
 #Check if GSL is installed
 if(USE_GSL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,9 @@ if(CMAKE_BUILD_TYPE MATCHES "Debug")
 endif(CMAKE_BUILD_TYPE MATCHES "Debug")
 
 #------------------------------------------------------------------------------
+#We are going to include this additional module here since it could be useful
+#when setting all of the following options.
+include(CMakeDependentOption)
 
 #Install options
 option(BUILD_ACQ "Build and install Acquisition software" ON)


tking53 (approved):
properly finds gsl-config on kqxhc, both the default gsl-1.15 and the module loaded gsl-2.3. also works on my laptop under the same conditions.

ksmith0:
@tking53 Does it properly handle when gsl is not installed? What happens if `USE_GSL=OFF`?

spaulaus:
The `USE_GSL` flag defaults to `OFF` and cannot be turned on by the user since its listed as dependent on another flag (`BUILD_UTKSCAN`). It is only turned on when `BUILD_UTKSCAN = ON`

ksmith0:
Does CMake Dependent Variable work for CMake 2.8.8 and newer?

tking53:
the `USE_GSL` flag shows up in ccmake (under advanced) when `BUILD_UTKSCAN` is turned on. If i then force it off it breaks like it used too. but it defaults to ON like stan mentioned and is hidden from the short list. Not sure about 2.8.8 but it works on kqxhc's default cmake (2.8.12.2). Do you have a computer with 2.8.8 that i could test it on?

spaulaus:
@ksmith0 I believe so. I will confirm for you.

spaulaus:
CMakeDependentOption was added to the CMake repository on October 19, 2006. Version 2.4.4 was released on November 21, 2006. I would assume that it's been part of the CMake distribution for 2.4.4+

spaulaus:
@ksmith0 How do you want to handle the situation where the user enables UTKSCAN and then manually disables GSL through the advanced options? 

If UTKSCAN is enabled and it cannot find GSL then the build will fail, which is what we want. 

We cannot stop users from adjusting advanced options. This isn't really a bug in my opinion If you have a suggestion I'm open to revising it.

ksmith0:
Let's just mark it as `INTERNAL` and be done with it. Then the user can't switch it at all.

spaulaus:
Sounds reasonable. Will do.

spaulaus:
If I set the option as internal then CMakeDependentOption will no longer work....

spaulaus:
@ksmith0  

I'm trying to set the option to internal, but am running into an issue. If I set it as internal then CMakeDependent option no longer works as expected. I have attempted the following: 

```cmake
CMAKE_DEPENDENT_OPTION(USE_GSL "Compile with GSL" ON "BUILD_UTKSCAN" OFF)
set(USE_GSL ${USE_GSL} CACHE INTERNAL "Compiles with GSL")
```
AND
```cmake
set(USE_GSL OFF CACHE INTERNAL "Compiles with GSL")
CMAKE_DEPENDENT_OPTION(USE_GSL "Compile with GSL" ON "BUILD_UTKSCAN" OFF)
```

Neither of these two cause the desired functionality.

ksmith0:
Do we even need `CMAKE_DEPENDENT_OPTION`? You can just move the `USE_GSL` flag down into Utkscan. Or move it into the if statement like so:
```cmake
if(BUILD_UTKSCAN)
   option(USE_GSL "Use GSL for Pulse Fitting" OFF)

   #Check if GSL is installed
   if(USE_GSL)
      find_package(GSL REQUIRED)
      add_definitions("-D usegsl")
   endif(USE_GSL)

   add_subdirectory(Utkscan)
endif(BUILD_UTKSCAN)
```

Then the option only appears once `BUILD_UTKSCAN` is enabled. Although once we make it internal it won't appear ever.

spaulaus:
That's not possible since we require GSL for some of the Resources along with their unit tests. Right now the option only appears if BUILD_UTKSCAN is enabled, it's essentially the same functionality.

ksmith0:
Oh well, then we can't have `USE_GSL` dependent on `BUILD_UTKSCAN` then.

spaulaus:
At the moment these Resources are only useful when using utkscan since nobody else uses the GSL for anything. If other programs began using GSL then I would agree that this should be changed.

ksmith0:
I suggest adding this to `Analysis/Utkscan/CMakeLists.txt` the current dev branch:

```cmake
if (NOT USE_GSL)
   message(FATAL_ERROR "GSL is required to build Utkscan!")
endif (NOT USE_GSL)
```

spaulaus:
I think that's a good compromise right now.

PR #220, opened 2017-02-06
 I changed the MAX_FILE_SIZE defintion (Line 55) from 4GB (4294967296…

sburcher:
I changed the definition of MAX_FILE_SIZE from 4 GB to 2GB. I tested the change on r230 with a pulser at  ~100 kHz and trace length set to 0.400.  A new ldf was created once the file was large enough that the next spill + EOF buffer would have put it over 2 GB.  

I also added a comment line in the write_data() function which explains where the number 65552 comes from.


ksmith0 (approved):
File rolls over before hitting 2GB.

Issue #148, opened 2016-12-08
MCA from poll2 in pacman mode


ksmith0:
MTAS group asked for the capability to add MCA mode to poll2 while in pacman mode. I do not see an issue with this as mca is already blocked if the acquisition is running. The only foreseeable issue is that the stop command is block and the mca may be set for a long run with no easy way to end it.

tking53:
implemented in #221

PR #221, opened 2017-02-06
MCA in pacman mode (issue #148)

sburcher:
The code to run the "mca" command inside poll2_core.cpp was contained
in an if statement if(!pac_mode).  I simply moved the chunk refering to
mca to the outside (just before) the if(!pac_mode). I tested it in both
pacman mode and not pacman mode and it seemed to run fine. (Addresses Issue #148)

### DIFF
--- a/Acquisition/Poll/source/poll2_core.cpp
+++ b/Acquisition/Poll/source/poll2_core.cpp
@@ -1441,6 +1441,36 @@ void Poll::CommandControl(){
 				debug_mode = true;
 			}
 		}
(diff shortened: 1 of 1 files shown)


ksmith0 (approved):
Looks like a straight forward move. I believe this should be fine, but I have not tested it.

tking53:
quick rebase before testing it

tking53 (changes requested):
mca is not in the poll2 help output when in pacman mode. (its there for normal mode so we could probably just copy the line)

tking53 (approved):

PR #222, opened 2017-02-07
Adding additional check on GSL in the utkscan CMakeLists

spaulaus:
The GSL flag can be disabled by an advanced user. This would not throw an error until the linking stage. The new addition will catch this error much sooner and provide the user with some information about what happened.


ksmith0:
Discussion about this addition is in #219

tking53 (approved):
This works as expected. If an advanced user forces `USE_GSL` to off with `BUILD_UTKSCAN` set to ON cmake will not complete and will complain with a helpful error message.

PR #223, opened 2017-02-07
Resolves Issue #106, #2, #16 and #87 - Rewriting ChannelData, ChannelEvent, etc

spaulaus:
This pull request addresses issues #106, #2, #16 and #87

All of the data classes are now derived from the XiaData class either directly (Processed XiaData) or through several generations. We no longer need a bunch of wrapper functions in HighResTimingData, or any of the other data classes since we are able to access any of the data through the methods defined by the base classes.

All of the updates to the various processors was to ensure that these data were being accessed properly and for cleanup.

During testing it became necessary to have the XiaDataDecoder class store not only the arrival time of the signal but also the Time sans CFD information. Offline timing needs the filter time sans CFD information.


Analysis/Utkscan/core/source/BarBuilder.cpp
```diff
@@ -26,14 +26,14 @@ void BarBuilder::BuildBars(void) {
 	    TimingDefs::TimingIdentifier key =
 	     	make_pair(it->first, list_.at(it->second)->GetChanID().GetSubtype());
 	    hrtBars_.insert(make_pair(key,
-            BarDetector(HighResTimingData(list_.at(it->second)),
-                        HighResTimingData(list_.at(mate->second)), key)));
+            BarDetector(HighResTimingData(*list_.at(it->second)),
+                        HighResTimingData(*list_.at(mate->second)), key)));
```
ksmith0: Can we rewrite this to:

```c++            
BarDetector(HighResTimingData(*(list_.at(it->second))),
                     HighResTimingData(*(list_.at(mate->second))), key)));

```

ksmith0 (approved):
Reviewed on:
```
$ uname -a && gcc --version && cmake --version
Linux step2f25 4.9.3-200.fc25.x86_64 #1 SMP Fri Jan 13 01:01:13 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)
Copyright (C) 2016 Free Software Foundation, Inc.

cmake version 3.6.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

Tested with HAGRiD data acquired with r230 using firmware R36101. The display of traces and fitting worked as expected.

Issue #224, opened 2017-02-09
If XIA or PLX isn't installed then the build will fail while searching for the missing option


spaulaus:
The logic to automatically toggle `BUILD_SUITE` to off was slightly flawed. The original logic looked like this 
```cmake
if (BUILD_ACQ)
	if(NOT BUILD_ACQ_ATTEMPTED AND NOT PLX_FOUND AND NOT XIA_FOUND)
		set (BUILD_ACQ OFF CACHE BOOL "Build and install PixieSuite" FORCE)
        else (PLX_FOUND OR XIA_FOUND)
...
```
This would cause the build to fail if the user has only one of the two packages installed (either PLX or XIA). You can consider that users may install PLX if they have other PLX devices on there system. Alternatively, the users could want only XIA if they're working on something related to set files, var files, etc. 

The logic should be amended to the following
```cmake
if (BUILD_ACQ)
	if(NOT BUILD_ACQ_ATTEMPTED AND NOT (PLX_FOUND AND XIA_FOUND))
				FF CACHE BOOL "Build and install PixieSuite" FORCE)
        else (PLX_FOUND OR XIA_FOUND)
...
```
This prevents the error message from occurring.

ksmith0:
I would not call it flawed, the original intent was to have this functionality as it was thought that the user would only install these packages if they were using the ACQ.

PR #225, opened 2017-02-09
Issue224 - XIA or PLX isn't installed then the build will fail

spaulaus:
This was causing the build to fail if only one of XIA or PLX was installed. The issue here is that the user may want only one or the other of these programs installed for a variety of reasons. We hadn't considered this possibility originally. The need for this has recently arisen and we've needed to adjust our logic here.


ksmith0 (approved):
Reviewed on:
```
$ uname -a && gcc --version && cmake --version
Linux step2f25 4.9.3-200.fc25.x86_64 #1 SMP Fri Jan 13 01:01:13 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)
Copyright (C) 2016 Free Software Foundation, Inc.

cmake version 3.6.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

BUILD_ACQ is disabled if either PLX or XIA libraries are not found.

ksmith0:
This resolved issue #224

Issue #216, opened 2017-02-06
Prepend CMake variables with PAASS_


ksmith0:
Some of the CMake variable are hard to find in curses mode. If they were all prepended with `PAASS_` they would at least be grouped together.

spaulaus:
I'll take care of this once Pull Request #223 is closed out. There's some overlap with some of the CMakeLists changes in there. Plus, I've already finished it.

PR #226, opened 2017-02-12
Issue #216 - Updated cmake options to be prepended with PAASS_.

spaulaus:
This commit resolves issue #216. This will make finding the custom PAASS related options easier in cmake. This is similar to what's done with GEANT4.


tking53:
I think that the `PAASS_USE_NCURSES` needs to stay hidden in `ccmake ` since the code will not compile with it off and the errors you get are not obvious. 

I know this is not related to the issue but might be helpful. Can you bring the `UNITTESS++_*` flags up to the short list when the user turns on `PAASS_BUILD_UNITTESTS` since it doesn't always find the paths if unittest-cpp is not installed in a "normal" location.

spaulaus:
> I know this is not related to the issue but might be helpful. Can you bring the UNITTESS++_* flags up to the short list when the user turns on PAASS_BUILD_UNITTESTS since it doesn't always find the paths if unittest-cpp is not installed in a "normal" location.
What do you mean by a "normal" location? For example, are you installing to ${HOME}? If so, then I would argue that you're an advanced user and should be able to update those paths as necessary. We cannot cover every possibility that users may encounter here.

tking53:
I know we cant find it everywhere but it would be nice to not have to sort thru the "advanced" list to find them. So if we bring them up to the "basic" list, it would be helpful.

ksmith0:
The nurses flag issue is a good point, I think this should be internal actually. 

I agree with stan on Unittest stuff. Only developers will use Unitest++ and should know about the advanced flags.

spaulaus:
I'm looking into the ncurses flag now.

spaulaus:
It looks like the only line that I missed prepending with `PAASS_` was the one that marked NCURSES as advanced.

spaulaus:
@tking53 The issue is now fixed. Good catch.

ksmith0:
Although I could agree with Toby on the Unittest flags as well. If someone selects to build the tests they most likely wouldn't mind the additional flags.

spaulaus:
I think leaving them as advanced is probably the best course here. They are not something that we need to clutter up a standard user's interface with.

tking53 (approved):
utkscan, unittests + defaults all build, scope runs as expected. Scan opens (with the svp/testing.xml cfg). Both on my laptop (Ubuntu16.10 kernel 4.8.0-37-generic gsl2.1 gcc 6.2.0-5ubuntu-12) and on kqxhc (system defaults + gsl1.15)

Issue #227, opened 2017-02-13
The XML files need to have better error handling and XML Parsing needs consolidated


spaulaus:
# Description of the problem
Right now we have little to no error handling on the utkscan configuration file. We currently have 5 separate files that read the configuration, and three of those five read the same node multiple times. This also breaks a fundamental idea that a class should have just one function. Therefore, we need to remove the XML reading functionality from the classes to help ensure that they do not take on too much responsibility. Here is the breakdown of the current nodes in the configuration and where they are read

| Name | Read From | 
|:-------:|:-----------:|
| Description | Globals::Globals |
| Global | Globals::Globals |
| Reject | Globals::Globals |
| Physical | Globals::Globals |
| Trace | Globals::Globals |
| Cfd | Globals::Globals |
| Fitting | Globals::Globals |
| Map | DetectorLibrary::LoadXml |
| TreeCorrelator | DetectorLibrary::LoadXml |
| DetectorDriver | DetectorDriver::LoadDrivers |
| Map | DetectorDriver::ReadCalXml |
| Map | DetectorDriver::ReadWalkXml |
| Notebook | Notebook::Notebook |
| TreeCorrelator | TreeCorrelator::buildTree |
| TimeCalibrator | TimeCalibrator::ReadTimingCalXml |

# Proposed Solution
I propose that we create an XmlInterface and XmlParser in Resources. This provides us with some base classes that can be used to implement XML reading for any of the programs not just utkscan. 
## XmlInterface
The XmlInterface will be a singleton that does the following operations:
* Opens a provided XML file
* Provides a pointer to the opened document
* Closes the document upon destruction

## XmlParser
The XmlParser class provides basic functionality of parsing various nodes. This class also provides some error handling methods that can be used in the children since they will all need some kind of verification. 

## Cleaning up of the Globals class
The Globals class has historically had a lot of information in it that it did not actually need. For example it had physical constants, type definitions and functions. All of these will be moved to the appropriate places. This class will now at it's core just be a way to access information that is stored in the XML file.

spaulaus:
# What we tested during the review
## Removed closing bracket from Configuration node. 
### Result: (PASS)
The Pugixml library caught the error and the code exited cleanly
## Commented out Revision Node
### Result : (PASS)
The Error handling on the Globals parsing caught the error and the program exited cleanly. 
## Commented out Globals node 
### Result: (PASS)
The code caught the error when checking the root node integrety, and the code exited cleanly. 
[...]

Issue #64, opened 2016-08-29
HisDrr produced files are not always readable by DAMM


ksmith0:
@spaulaus reported this issue.

ksmith0:
Are these the HisDrr files from MCA or ScanLib?

spaulaus:
ScanLib is the only time we've really seen this issue. We haven't undergone thorough testing to figure out what's causing this issue.

spaulaus:
I again think that this isn't a huge issue. The files are always readable by versions of DAMM that are compiled on the computer.

PR #229, opened 2017-02-21
Resolves Issue #227 - XML files need to have better error handling

spaulaus:
This pull request resolves Issue #227. This is a nearly complete overhaul to the handling of XML files. There are new classes that are dedicated to parsing specific nodes of the configuration file. For example, the Map node is now parsed once from a single file instead of three times from two files. The Globals class has been gutted of any kind of XML parsing, it also has the constants and functions removed from the header.

I have now added a XmlInterface and XmlParser, to the resources. These facilitate other programs to use XML files if they choose.


Analysis/Utkscan/core/source/TimingCalibrator.cpp
```diff
@@ -32,7 +32,7 @@ TimingCalibrator* TimingCalibrator::get() {
 }
 
 void TimingCalibrator::ReadTimingCalXml() {
```
mvostinar, line 33: This should follow the same programing as the Notebook since it is reopening the same .xml file again.

mvostinar (dismissed):
There is an issue which needs to be addressed, see the comment in the file. Otherwise the code seems fine.

Analysis/Utkscan/core/source/TimingCalibrator.cpp
spaulaus: I updated this to follow the same programming as in Notebook. This class needs to be updated to handle the geometric correction a little differently, but this will bring it in line with the scope of this pull request.

mvostinar (approved):
Ubuntu version 14.04.1 -#64
Cmake version 3.7.1
gcc version 4.8.4

It passed all the testing that was reported in the issue #227. The program was correctly issuing the error messages when made on purpose. It was behaving in the expected way for all the tests. One additional side possible issue has been find which can be addressed in the future.

Issue #230, opened 2017-02-22
XiaData::GetBaseline() needs documentation clarification


tking53:
There is no indication in the code that XiaData::GetBaseline() requires options enabled on the modules and that we should (in general) use the HighResTimingData::GetAvgBaseline() which is calculated in the analyzers.

spaulaus:
Users should not "in general" use one or the other. They should use the method that is appropriate to their specific needs.

tking53:
Resolved by #231

PR #231, opened 2017-02-22
Add clarification to the 2 functions that return baseline information

tking53:
This should fix issue #230.

XiaData::GetBaseline() and HighResTimingData::GetAvgBaseline() are the 2 functions

### DIFF
--- a/Analysis/ScanLibraries/include/XiaData.hpp
+++ b/Analysis/ScanLibraries/include/XiaData.hpp
@@ -89,9 +89,10 @@ public:
     ///@return True if this channel was generated on the module
     bool IsVirtualChannel() const { return isVirtualChannel_; }
 
-    ///@return The baseline as it was calculated on the module.
-    ///The calculation was done using a filter.
-    /// This requires that ESUMS parameter be enabled on the modules during acquisition.
+    ///@return The baseline as calculated on-board the Pixie-16 modules using
+    /// the energy filter. This parameter is only set if the data set
+    /// contains the Energy Sums in the list mode data. This baseline cannot
+    /// be used in conjunction with trace information.
     double GetBaseline() const { return baseline_; }
 
     ///@return The energy that was calculated on the module
--- a/Analysis/ScanLibraries/include/XiaData.hpp
+++ b/Analysis/ScanLibraries/include/XiaData.hpp
@@ -89,7 +89,9 @@ public:
     ///@return True if this channel was generated on the module
     bool IsVirtualChannel() const { return isVirtualChannel_; }
 
-    ///@return The baseline as it was calculated on the module
+    ///@return The baseline as it was calculated on the module.
+    ///The calculation was done using a filter.
+    /// This requires that ESUMS parameter be enabled on the modules during acquisition.
     double GetBaseline() const { return baseline_; }
 
     ///@return The energy that was calculated on the module
--- a/Analysis/Utkscan/core/include/HighResTimingData.hpp
+++ b/Analysis/Utkscan/core/include/HighResTimingData.hpp
@@ -42,13 +42,17 @@ public:
         return(false);
     }
 
-    /** \return The current value of aveBaseline_ .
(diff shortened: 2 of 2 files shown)


Analysis/ScanLibraries/include/XiaData.hpp
```diff
@@ -89,7 +89,9 @@ class XiaData {
     ///@return True if this channel was generated on the module
     bool IsVirtualChannel() const { return isVirtualChannel_; }
 
-    ///@return The baseline as it was calculated on the module
+    ///@return The baseline as it was calculated on the module.
```
spaulaus: The notes about what the function returns shoudl be above this line. The units here are in trapezoidal filter units, whatever they are. Make sure that you note that the baseline that's provided here cannot be used in conjunction with the QDC or subtracted from the trace as we do with the baseline calculated from the trace.

Analysis/Utkscan/core/include/HighResTimingData.hpp
```diff
@@ -41,7 +41,8 @@ class HighResTimingData : public ChanEvent {
         return(false);
     }
 
-    /** \return The current value of aveBaseline_ */
+    /** \return The current value of aveBaseline_ .
+      This value is calculated during the fitting procedure as a straight sum, from the trace. */
```
spaulaus: The note about this should come before the return. Also, the baseline average is not a "straight sum", this may mislead users to think that we are simply calculating a QDC. Maybe update the copy to read

> This baseline is calculated from the trace by averaging the points preceding the waveform.

spaulaus (changes requested):
The notes as listed do not provide the proper information to the users. In some cases they may increase the uncertainty the users may have.

Analysis/ScanLibraries/include/XiaData.hpp
tking53: How is this?
``` 
    ///The calculation was done using a trapezoidal filter.
    ///This requires that ESUMS parameter be enabled on the modules during acquisition.
    ///The baseline returned by this function can not be used with values calculated from the trace, like the QDC.
    ///It also can not be subtracted from the trace like we do with the calculated baseline (HighResTimingData::GetAveBaseline())
    ///@return The baseline as it was calculated on the module, in trapezoidal filter units.
    double GetBaseline() const { return baseline_; }
```

Analysis/Utkscan/core/include/HighResTimingData.hpp
tking53: And this?
```
    /** This baseline value is calculated from the trace by averaging the points preceding the waveform
        \return The current value of aveBaseline_ .*/
    double GetAveBaseline() const { return GetTrace().GetBaselineInfo().first; }
```

Analysis/ScanLibraries/include/XiaData.hpp
spaulaus: > ///The calculation was done using a trapezoidal filter.
    ///This requires that ESUMS parameter be enabled on the modules during acquisition.
    ///The baseline returned by this function can not be used with values calculated from the trace, like the QDC.
    ///It also can not be subtracted from the trace like we do with the calculated baseline (HighResTimingData::GetAveBaseline())
    ///@return The baseline as it was calculated on the module, in trapezoidal filter units.
    double GetBaseline() const { return baseline_; }

I would prefer
```c++
///@return The baseline as calculated on-board the Pixie-16 modules using the 
/// energy filter. This parameter is only set if the data set contains the Energy 
/// Sums in the list mode data. This baseline cannot be used in conjunction with 
///trace information. 
```

Also, make sure you always conform to the 80 character line limits. See the [Style Guidelines](https://github.com/pixie16/paass/wiki/Style-Guidelines#formatting)

Analysis/Utkscan/core/include/HighResTimingData.hpp
spaulaus: That should be fine, just make sure you adhere to the 80 char limit.

spaulaus (approved):

Issue #232, opened 2017-02-23
RootFitter does not do anything useful and needs updated


spaulaus:
# Description
The RootFitter class doesn't do anything useful. Development was stalled out on it since there was some kind of funky ROOT related compilation issue. This issue has since been resolved and this fitter should be fixed. Implementation of this fitter along with the additional Function classes provides programs that do fitting (utkscan, scope) a consistent functional framework to work from. 

# Work to be completed
* Update RootFitter so that it produces viable results
* Update unittest-RootFitter so that it tests all of the failure modes that the RootFitter can return.
* Update FittingAnalyzer so that it has an option for using the RootFitter class.
* Update FittingAnalyzer so that it still compiles if ROOT is not found in the system.
* Scope uses the VandleTimingFunction for fitting. 

# Acceptance Criteria
* Successful compilation if  ROOT is not found in the system but GSL is found. 
* RootFitter class passes all of the unittests.
* RootFitter analysis results are not statistically different than results from GslFitter.

spaulaus:
# Test data information
I analyzed `kqxhc:/data/utk/pixieworkshop/pulser_003.ldf`. The configuration file can be found [here](https://github.com/pixie16/paass/blob/dev/Analysis/Utkscan/share/utkscan/cfgs/svp/pixieworkshop.xml). I made minor modifications to the configuration file to reflect some updates to the software.  

# Results using GslFitter (reference histogram)
The FWHM of the Time Difference was 33.7208 ps
![image](https://cloud.githubusercontent.com/assets/1553042/23277697/d428e8ec-f9dc-11e6-8580-4db07ea8a2c5.png)

# Results using RootFitter
The FWHM of the Time Difference was 33.7244 ps
![image](https://cloud.githubusercontent.com/assets/1553042/23314069/5382a856-fa8e-11e6-979b-868539b27ff7.png)

# Processing times
The scan spends on average (20 samples) 52% more time in the FittingAnalyzer when we are using the RootFitter than the GslFitter. This result isn't surprising. The GslFitter uses predefined functions and derivatives to perform the fit, where as ROOT uses numerical methods to calculate the derivatives. Another reason is that we are setting up a TGraph before each fit. The overhead of creating the TGraph at each iteration may be causing some speed issues as well.

# Conclusions
The RootFitter will work for analyzing signals. It only supports the VandleTimingFunction.

PR #239, opened 2017-02-28
Resolves Issue #232 - RootFitter does not do anything useful

spaulaus:
This fixes Issue #232. The RootFitter has been updated to so that it produces accurate results. The scope has been updated to use the proper fitting functions.

### DIFF
--- a/Analysis/Resources/include/RootFitter.hpp
+++ b/Analysis/Resources/include/RootFitter.hpp
@@ -5,12 +5,9 @@
 #ifndef _PIXIESUITE_ROOTFITTER_HPP_
 #define _PIXIESUITE_ROOTFITTER_HPP_
 
-#include <TF1.h>
-#include <TFitResult.h>
-#include <TGraphErrors.h>
-
 #include "TimingDriver.hpp"
 
+class TF1;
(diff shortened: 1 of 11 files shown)


spaulaus:
# Testing
## Program compiles if ROOT is not present. 
### Result : PASS
The program compiles and links properly when ROOT is not installed. The GSL 1 fitter was throwing fits, but the program didn't have any major issues. 

## Scope fits and plots in bins
### Result : PASS
The scope fits the data and provides results that are consistent with what we expect. The plot and fit results are displayed in bins. We tried to break the fit by providing bogus values to the range but things worked without issue. E.x. `fit 20 0`, `fit 20 -30`, `fit 20 70`. The traces were less than 70 bins long. 

## Tested results from utkscan
### Result : PASS
The results from the test data file showed no statistical deviation from the GslFitter. 

## Ran the unittest
### Result : PASS

mvostinar (approved):
Ubuntu version 14.04.1 -#64
Cmake version 3.7.1
gcc version 4.8.4

Tested the unittest, the scope and the utkscan and the program behaved normally. The gsl fitter problem was reported in issue #240.

Issue #234, opened 2017-02-27
Analysis/Utilities folders should be cleaned up


spaulaus:
# Description
The Analysis/Utilities folder is a little bit of a mess. The programs contained here should be reorganized so that they each have their own directory. This will keep the software organized as the number of Utility programs grows. 

# Work to be completed
1. Create a new directory in the Utilities directory for each of the Utility programs
2. Move all of the utilities into their dedicated folders
3. Update cmake files to reflect the new directory structure. 

# Acceptance Criteria
1. All of the programs compile as before.
2. Each program is in its own folder.

PR #241, opened 2017-03-01
Resolves Issue #234 : Analysis/Utilities folders should be cleaned up

spaulaus:
This resolves issue #234. The Analysis/Utilitiy directory was reorganized so that the software now all live in their own directories. This keeps the software more organized and prevents confusion over what header should belong with what source file. This organization is more extensible as the number of utility programs begins to grow. 

The traceFilterer program is not included in the compilation since it wasn't include before. The program should be revised to fix the compilation issues.


mvostinar (commented):
Managed to build and make it without any issues. The headReader, scope and skeleton can be found in the install/bin directory, however the TraceFilterer is not there. I tried to see if I have missed to turn on something when the make is done, but I can't see anything. I just wanted to check if the TraceFilterer should behave the same as the rest of the programs? (According to the code it should, but I don't know precisely what it is supposed to do.)   

I have just read the last sentence in your comment and saw that it is not included. Sorry.

mvostinar (approved):
Ubuntu version 14.04.1 -#64
Cmake version 3.7.1
gcc version 4.8.4

The build and make worked without an issue. The headReader, Scope and skeleton are now in separate programs. When run they behave normally.

Issue #240, opened 2017-02-28
Gsl1Fitter has issues with segmentation faults


spaulaus:
# Desicription
The GSL v. 1 fitter is segfaulting while trying to analyze traces. This issue has been reported by multiple users. The GSL fitters should have additional error handling updated. 

# Work to be performed
1. Identify the cause of the GSL v1 segmentation faults
2. Remedy the issues that were found

# Acceptance Criteria
1. The code compiles
2. Data can be analyze and produce results that are statistically consistent with other fitting methods.

spaulaus:
Analyzed data with the GSL1Fitter class. Used pulser_003.ldf from the pixieworkshop. 
The FWHM of the Time Difference was 33.7114 ps

![image](https://cloud.githubusercontent.com/assets/1553042/23484350/9bce69be-fea4-11e6-8d94-15717ccf1741.png)

PR #243, opened 2017-03-01
Resolves Issue #240 - Gsl1Fitter has issues with segmentation faults

spaulaus:
This resolves issue #240. 

The vectors that hold the data and weights was not being deleted before returning. This caused the segmentation fault on the second time through the Analyze method. 

In addition, the vector that held the fitted parameters was being freed before we actually got the parameter from it. This should resolve all the known issues with the Gsl1Fitter class.

### DIFF
--- a/Analysis/Resources/source/Gsl1Fitter.cpp
+++ b/Analysis/Resources/source/Gsl1Fitter.cpp
@@ -51,11 +51,12 @@ double GslFitter::CalculatePhase(const std::vector<double> &data,
                            const std::pair<double, double> &pars,
                            const std::pair<unsigned int, double> &max,
                            const std::pair<double, double> baseline) {
+
     gsl_multifit_function_fdf f;
     int status;
     const size_t sizeFit = data.size();
     size_t numParams;
-    double xInit[3];
+    double xInit[2];
 
     double *y = new double[sizeFit];
     double *sigma = new double[sizeFit];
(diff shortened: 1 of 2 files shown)


tking53 (changes requested):
Code compiles on kqxhc (see below). However it fails its unittest. 
```
/home/tking/programs/paass/Analysis/Resources/tests/unittest-GslFitter.cpp:25:1: error: Failure in TestGslFitter: Expected 0.85658 +/- 0.1 but was 6.90692e-310
FAILURE: 1 out of 1 tests failed (1 failures).
Test time: 0.00 seconds.

```
```
Linux Flavor

Linux Kernel 
3.10.0-514.6.1.el7.x86_64 

cmake version 2.8.12.2 

Copyright (C) 2015 Free Software Foundation, Inc.

GSL version 
1.15
```

spaulaus:
@tking53 : I have removed the unit test's check on the amplitude. We do not really care if the amplitude is different than what we thought it was going to be. In addition, the Gsl1Fitter no longer sets the amplitude variable, since it was never used for anything in the analysis.

tking53 (approved):

tking53:
Code compiles and runs on kqxhc (See my comment above for environment) the Gsl1Fitter now runs. The code produces the vandle histograms (3242-3245). Compared against the Gsl2Fitter these histograms are very similar (small fluctuation in the gamma flash centroid bin for 3245 which is fixable with a recalibration).

PR #248, opened 2017-03-02
Resolves #247 - Bad logic in the GlobalsXmlParser::ParseRejectNode

### LINKED ISSUE #247: Error in GlobalsXmlParser::ParseRejectNode logic
# Description
There's a bug in the parsing of the rejection regions. The logic 

```c++
if ( start == 0 || end == 0 || start > end) {
            ss << "Globals: incomplete or wrong rejection region "
               << "declaration: " << start << ", " << end;
            throw invalid_argument(ss.str());
        }
```
Should read

```c++
if ( (start == 0 && end == 0) || start > end) {
            ss << "Globals: incomplete or wrong rejection region "
               << "declaration: " << start << ", " << end;
            throw invalid_argument(ss.str());
        }
```

# Work to be done 
1. Correct logic so that the rejection regions are read properly when the start is set to 0

# Acceptance criteria
1. Rejection regions with start=0 are read properly.

spaulaus:
The previous logic was not allowing users to parse rejection regions with a start of 0.

### DIFF
--- a/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
+++ b/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
@@ -220,7 +220,7 @@ vector<pair<unsigned int, unsigned int> > GlobalsXmlParser::ParseRejectNode(
         int end = time.attribute("end").as_int(0);
 
         std::stringstream ss;
-        if (start == 0 || end == 0 || start > end) {
+        if ( (start == 0 && end == 0) || start > end) {
             ss << "Globals: incomplete or wrong rejection region "
                << "declaration: " << start << ", " << end;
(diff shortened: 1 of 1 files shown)


tking53 (approved):
Code compiles on kqxhc, the code correctly parses the rejection region (as displayed during the initialization of the scan interface). The code also catches the 2 error scenarios (both equal to 0 or end time before start). A quick scan with the data shows that the regions appear to be used. 


```
Linux Flavor

Linux Kernel 
3.10.0-514.6.1.el7.x86_64 

cmake version 2.8.12.2 

Copyright (C) 2015 Free Software Foundation, Inc.

GSL version 
2.3

```

Issue #253, opened 2017-03-14
eventWidth is not applied to UtkUnpacker


rin-yokoyama:
# Description
eventWidth of [UtkUnpacker](https://github.com/pixie16/paass/blob/dev/Analysis/Utkscan/core/source/UtkUnpacker.cpp) is not loaded properly. It is initialized with the default value of [Unpacker](https://github.com/pixie16/paass/blob/dev/Analysis/ScanLibraries/source/Unpacker.cpp#L220) and uses this value regardless of the input config file.

# Work to be done
Set proper eventWidth value of UtkUnpacker from [Globals](https://github.com/pixie16/paass/blob/dev/Analysis/Utkscan/core/source/Globals.cpp).

# Acceptance Criteria
1. All of software compiles
2. Builds events with a proper time window given in the input config xml.

rin-yokoyama:
I send a pull request to this issue

ksmith0:
You can simply list the issue number with the pound symbol in the pull request (#254) to cross document.

tking53:
This was addressed in pull requests #254 and #257 . Closing the issue.

PR #254, opened 2017-03-14
UtkUnpacker.cpp was modified to load eventWidth from Globals

rin-yokoyama:
This will make UtkUnpacker to build events with given event width in an input config xml.

### DIFF
--- a/Analysis/Utkscan/core/source/UtkUnpacker.cpp
+++ b/Analysis/Utkscan/core/source/UtkUnpacker.cpp
@@ -60,6 +60,9 @@ void UtkUnpacker::ProcessRawEvent(ScanInterface *addr_/*=NULL*/) {
         PrintProcessingTimeInformation(systemStartTime, times(&systemTimes),
             GetEventStartTime(), eventCounter);
 
+    //set eventWidth from Globals (loaded from config file)
+    eventWidth = Globals::get()->GetEventLengthInTicks();
+ 
     if (Globals::get()->HasRejectionRegion()) {
         double eventTime = (GetEventStartTime() - GetFirstTime()) *
                                 Globals::get()->GetClockInSeconds();
--- a/Analysis/Utkscan/core/source/UtkUnpacker.cpp
+++ b/Analysis/Utkscan/core/source/UtkUnpacker.cpp
@@ -60,9 +60,6 @@ void UtkUnpacker::ProcessRawEvent(ScanInterface *addr_/*=NULL*/) {
         PrintProcessingTimeInformation(systemStartTime, times(&systemTimes),
             GetEventStartTime(), eventCounter);
 
-    //set eventWidth from Globals (loaded from config file)
-    eventWidth = Globals::get()->GetEventLengthInTicks();
- 
     if (Globals::get()->HasRejectionRegion()) {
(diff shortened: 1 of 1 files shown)


ksmith0:
Addresses issue #253

Analysis/Utkscan/core/source/UtkUnpacker.cpp
```diff
@@ -60,6 +60,9 @@ void UtkUnpacker::ProcessRawEvent(ScanInterface *addr_/*=NULL*/) {
         PrintProcessingTimeInformation(systemStartTime, times(&systemTimes),
             GetEventStartTime(), eventCounter);
 
+    //set eventWidth from Globals (loaded from config file)
+    eventWidth = Globals::get()->GetEventLengthInTicks();
```
ksmith0: This line should use the `Unpacker::SetEventWidth()` method, `eventWidth` is protected.

ksmith0 (dismissed):

Analysis/Utkscan/core/source/UtkUnpacker.cpp
ksmith0: Also, this should not be in `ProcessRawEvent()` as the command will be called repeatedly. I believe it should be placed in `UtkUnpacker::InitializeDriver()`.

ksmith0 (commented):
Looks good. I can not test it, so I can not approve the pull request.

tking53:
I have tested this. It works as expected and doesn't seem to break anything. It will require new Time Calibrations, but this is to be expected.  I tested this on kqxhc. Additionally, this one line change was added to my ornl2016 branch ( [https://github.com/tking53/paass/tree/ornl2016](url)  )  as well as Andrew Keeler's e14060 branch ( [https://github.com/akeeler/paass/tree/e14060](url) ) and it works there as well.

```
Linux Flavor

Linux Kernel 
3.10.0-514.6.1.el7.x86_64 

cmake version 2.8.12.2 

GSL version 
2.3

```

tking53 (approved):
Works as expected on kqxhc. See my comment for details.

Analysis/Utkscan/core/source/UtkUnpacker.cpp
spaulaus: Actually, this should happen even before that. I think it needs to go in [this ](https://github.com/pixie16/paass/blob/master/Scan/utkscan/core/source/UtkScanInterface.cpp#L64) method. Putting it into `UtkUnpacker::InitializeDriver()` is too late since it's already started to build events.

poecilorhyncha: You mean the first event is already build when UtkUnpacker::InitializeDriver() called? I will try initializing [there](https://github.com/pixie16/paass/blob/master/Scan/utkscan/core/source/UtkScanInterface.cpp#L64) then.

rin-yokoyama: You mean the first event is already build when UtkUnpacker::InitializeDriver() called? I will try initializing [there](https://github.com/pixie16/paass/blob/master/Scan/utkscan/core/source/UtkScanInterface.cpp#L64) then.

PR #257, opened 2017-03-21
Issue 253

rin-yokoyama:
According to the [comment](https://github.com/pixie16/paass/pull/254/files/e84ac80b868c2aa6eb1b692d2b51c1340827cb13), I modified the merged pull request #254.
UtkUnpacker::InitializeDriver() is called at [the first event loop](https://github.com/pixie16/paass/blob/dev/Analysis/Utkscan/core/source/UtkUnpacker.cpp#L58) so the fist event was not build with proper time window in #254.
I checked this modification worked fine with the akeeler's e14060 branch.

### DIFF
--- a/Analysis/Utkscan/core/source/UtkScanInterface.cpp
+++ b/Analysis/Utkscan/core/source/UtkScanInterface.cpp
@@ -94,6 +94,8 @@ bool UtkScanInterface::Initialize(string prefix_) {
 
     Globals::get()->SetOutputFilename(GetOutputFilename());
     Globals::get()->SetOutputPath(GetOutputPath());
+    //set eventWidth from Globals (loaded from config file)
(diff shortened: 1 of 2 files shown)


tking53:
This works as expected. I ran this on my laptop with my ornl2016 branch (ubuntu 16.10, kernel 4.8.0-42, gsl 2.3, gcc 6.2-ubuntu).

tking53 (approved):
as per Stan's suggestion. This works, I added it to my ornl2016 branch and it works (just like the previous fix) and Rin tested it with andrew's e14060 branch. See my above review to my system specifics.

UPDATE: This required me to do a manual rebase on my laptop since github didnt want to. The rebase was painless with just a quick re-push.

ksmith0:
Properly addressed #253.

Issue #256, opened 2017-03-16
ncurses compile issue 


ksmith0:
An ncurses compile issue has appeared when using  gcc version 4.8.2 20140120 (Red Hat 4.8.2-15) and cmake version 2.8.12.2. As shown by the following error:

```
/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld: ../../../Core/source/libPaassCoreStatic.a(CTerminal.cpp.o): undefined reference to symbol 'halfdelay'
/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld: note: 'halfdelay' is defined in DSO /lib64/libtinfo.so.5 so try adding it to the linker command line
/lib64/libtinfo.so.5: could not read symbols: Invalid operation
collect2: error: ld returned 1 exit status
make[2]: *** [Acquisition/Poll/source/poll2] Error 1
make[1]: *** [Acquisition/Poll/source/CMakeFiles/poll2.dir/all] Error 2
make: *** [all] Error 2
```

This can be resolved by add the following line after `find_package(Curses REQUIRED)`
```cmake
list(APPEND CURSES_LIBRARIES -ltinfo)
```

spaulaus:
@ksmith0 : 

I do not see this problem with gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609 and CMAKE 3.7.1

ksmith0:
I've never seen it except on bill.

spaulaus:
Another issue seen only on a single system.

PR #258, opened 2017-03-25
Resolves Issue 256: Adds a missing ld flag  (-ltinfo)

tking53:
This was needed on the Bill laptop. I would like Miguel to test this. (I cant find his github username in the list). This doesn't seem to break anything on my laptop (ubuntu 16.10 gsl2.3 gcc 6.2)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ find_package (XIA)
 #Find curses library used for pretty much everything at this point.
 if(PAASS_USE_NCURSES)
 	find_package(Curses REQUIRED)
+	list(APPEND CURSES_LIBRARIES -ltinfo)
 endif(PAASS_USE_NCURSES)
 
 if (CURSES_FOUND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,12 +94,26 @@ find_package (XIA)
 #Find curses library used for pretty much everything at this point.
 if(PAASS_USE_NCURSES)
 	find_package(Curses REQUIRED)
-	list(APPEND CURSES_LIBRARIES -ltinfo)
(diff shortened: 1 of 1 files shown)


ksmith0:
This is related to issue #256. I believe it is an issue with the compiler on that machine and should not be added to the cmake scripts as it should be automatically determined.

ksmith0:
Maybe something along these lines?

```cmake
CHECK_LIBRARY_EXISTS("${CURSES_NCURSES_LIBRARY}"
    cbreak "" CURSES_NCURSES_HAS_CBREAK)
  if(NOT CURSES_NCURSES_HAS_CBREAK)
    find_library(CURSES_EXTRA_LIBRARY tinfo)
    list(APPEND CURSES_LIBRARIES CURSES_EXTRA_LIBRARY)
  endif(NOT CURSES_NCURSES_HAS_CBREAK)
```

tking53:
I have implemented something similar to what @ksmith0 suggested. This works on the BILL laptop and correctly finds the needed library. it still compiles on kqxhc and my ubuntu laptop.

tking53:
i have run into a weird issue. where running just `cmake ../` (with flags for building utkscan and USE_HRIBF) then make does not build however if you use ccmake for the setup it will make without issue. We are currently investigating. Im going to leave this open so we can track the ideas for this.

ksmith0:
I will try to have a look at this tonight.

spaulaus (commented):
Since this occurs on only a single platform. I would rather we not include this in the code base.

spaulaus:
I'm closing this request since #256 has been closed as "won't fix". I will leave the branch so that the patch can be viewed if necessary.

PR #261, opened 2017-04-25
Added a save command to poll2.

ksmith0:
The save command will write the DSP parameters to the set file specified
in pixie.cfg or provides an option to write to a different file for
backup.


tking53:
it was missing a terminating "; that my laptop didn't care about but Takk bailed during compilation

tking53 (approved):
the save command works as expected. however the help wasn't detailed enough for Miguel. so i reworded that, and in the process of that i found that `save help` would print the expanded help message but would also save the parameters to a file called `help`. I fixed this. This was tested on takk

PR #265, opened 2017-05-11
Issue264: Readded QDC compression to the VANDLE histograms

### LINKED ISSUE #264: QDC compression is never used
The QDC compression value is parsed from the config file but is never actually used. Robert requested that it be re-implemented during the IS632 run

tking53:
This should resolve #264 

This was requested during the last experiment (IS632). This was picked off the IS632 branch and cleaned up. We had no issues with this on both the Bill laptop and kqxhc (offline scans).

### DIFF
--- a/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
+++ b/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
@@ -283,10 +283,14 @@ void GlobalsXmlParser::ParseTraceNode(const pugi::xml_node &node, Globals *globa
     else
         globals->SetQdcCompression(1.0);
 
-    sstream_ << "QDC Compression : " << globals->GetQdcCompression();
-    messenger_.detail(sstream_.str());
-    sstream_.str("");
-
+    if( globals->GetQdcCompression() == 0.0 )
+        throw invalid_argument(CriticalAttributeMessage("QDC compression can "
+                                                                "NOT be zero"));
+    else {
+        sstream_ << "QDC Compression : " << globals->GetQdcCompression();
+        messenger_.detail(sstream_.str());
+        sstream_.str("");
+    }
     if (!node.child("WaveformRange").empty()) {
         std::map<std::string, std::pair<unsigned int, unsigned int> > waveRngs;
         for (pugi::xml_node_iterator waveit = node.child("WaveformRange").begin();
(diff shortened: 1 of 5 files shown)


rin-yokoyama (approved):
It compiled both on kqxhc and my Surface.
It runs without any error at least on pspmt test data.

ksmith0:
I will not block this, but without testing with VANDLE data it does not make much sense to approve it. The change in functionality as I understand is only with regard to VANDLE and should not affect PSPMT data (although this is a good data point to have as well).

rin-yokoyama (changes requested):
There is no check that the value of qdcComp_ is not zero.
And I will test it with VANDLE data.

spaulaus:
> I will not block this, but without testing with VANDLE data it does not make much sense to approve it.

I **would** block this on the argument that this is added to the VandleProcessor. These histograms are meant to be "Raw" data. This compression is an experiment specific requirement and should not be forced into every situation. 

In addition, the addition of a dedicated variable is redundant. The parameter is read from the configuration file, which means that it can be accessed via Globals.

tking53:
I agree that its an experiment specific requirement, but that is why it defaults to 1 when its not present in the config. This allows us the flexibility to quickly have it if we need it. We ran the IS632 without a experiment specific processor per Miguel's wishes and we might continue this if possible.  
```
    if (!node.child("QdcCompression").empty())
        globals->SetQdcCompression(
                node.child("QdcCompression").attribute("value").as_double());
    else
        globals->SetQdcCompression(1.0);
``` 

@spaulaus I see that the dedicated variable is redundant, but i was being consistent with the multiplicity, the offset and the number of starts which are passed in by the DetectorDriverXmlParser. Is there a reason to reference Globals each time over the dedicated variable other than reducing the memory needed by the extra variable?

rin-yokoyama (approved):
I confirmed that it is compiled and runs on is632 data.
It compresses the qdc histogram properly and does not compress when there is no "QdcCompression" element in the config file.
It also throws an exception when the "value" attribute is zero or does not exist.

spaulaus:
> @spaulaus I see that the dedicated variable is redundant, but i was being consistent with the multiplicity, the offset and the number of starts which are passed in by the DetectorDriverXmlParser. Is there a reason to reference Globals each time over the dedicated variable other than reducing the memory needed by the extra variable?

I don't disagree that you could use the same procedure for the QDC compression. In fact, everything should have been updated so that the qdc compression was moved out of Globals and read in through the VandleProcessor constructor. This would enable different processors to have different compression factors. We should consider a new class (DammPlotManipulator?) that handles these variables without the redundancy involved.

PR #267, opened 2017-05-11
Resolves #266 (non-helpful error with a missing TraceDelay node)

### LINKED ISSUE #266: XML parser throws a non-helpful error with a missing TraceDelay node
When the TraceDelay node is not present in the config file, the error that is thrown is not very helpful. This is due to a missing `invalid_argument` call  in the GlobalsXmlParser.cpp 
```
throw CriticalNodeMessage(node.child("TraceDelay").name());
```
vs every other CriticalNodeMessage call in that file which look like this.
```
 throw invalid_argument(CriticalNodeMessage("WaveformRange"));
```

tking53:
Adds the missing  invalid_argument call, and resolves #266

### DIFF
--- a/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
+++ b/Analysis/Utkscan/core/source/GlobalsXmlParser.cpp
@@ -271,7 +271,8 @@ void GlobalsXmlParser::ParseTraceNode(const pugi::xml_node &node, Globals *globa
         globals->SetTraceDelay(
                 node.child("TraceDelay").attribute("value").as_uint());
     } else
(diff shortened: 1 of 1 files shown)


sztaylor89:
Error message is not helpful: 
"what():  XmlParser::ParseGlobal - We couldn't find the "" node. This node is critical to operation."

sztaylor89 (approved):
Works.  Builds and runs, and throws "what():  XmlParser::ParseGlobal - We couldn't find the "TraceDelay" node. This node is critical to operation." as correct error message

Issue #262, opened 2017-05-02
XiaListModeDecoder causes crash when headerLength == STATS_BLOCK


rin-yokoyama:
# Description
I was investigating the crash of utkscan on some Steven's data at ANL.
While scanning some of his data, utkscan suddenly start consuming memory significantly and was killed eventually.
I found that it was crashing at XiaListModeDecoder::DecodeBuffer().
Some lines were commented out in one of the switch cases, case STATS_BLOCK ([like this](https://github.com/pixie16/paass/blob/dev/Analysis/ScanLibraries/source/XiaListModeDataDecoder.cpp#L66)).
One of the lines, "buf+=eventLength", should not be commented out because the buf pointer stays at the same place and it loops forever creating [new XiaData instances](https://github.com/pixie16/paass/blob/dev/Analysis/ScanLibraries/source/XiaListModeDataDecoder.cpp#L43) otherwise.

The ANL1471 data seem to contain some blocks with headerLength == 1 (STATS_BLOCK).
I am not yet sure if those blocks in the data are actually the stats blocks or some kind of error blocks, but anyway the code should handle such unusual blocks without crashing.

# Work to be done
Modify XiaListModeDecoder to handle or skip a block with headerLength == 1 without causing a crash.

# Acceptance Criteria
1. All of software compiles
2. Should not loop forever in the while(true) statement of the DecodeBuffer() method.
3. Be able to analyze the rest of data behind the block.

ksmith0:
Was this data written with poll2 or pacman? You may want to consider scanning the data with a tool such as evt2root or hexdump to ensure that the data file is properly constructed.

sztaylor89:
It was written with pacman.  This was before poll2 existed yet.  @ksmith0 how would I go about checking the data with hexdump or evt2root?  are these built somewhere as part of paass?  What would I be looking for?

ksmith0:
The hexdump program is something that @cthornsb wrote and I'm not sure where it lives. 

I wrote [evt2root](https://github.com/ksmith0/evt2root), you should use version 3 (branch v3). The program is not capable of unpacking the XIA data, but will unpack the surrounding HRIBF binary format. You will want to use the evtDump program and pipe it to less then search for the word 'ERROR'.
```
evtDump -f hribf run0.ldf | less
```

ksmith0:
[...]

Issue #252, opened 2017-03-06
PAASS_BUILD_SCAN_UTILITIES does not do anything


tking53:
# Description 
The scan utilities (Scope, headreader, set2ascii and set2root) are still built even if this flag is off. I think this is just a missing if statement in the root CMakeLists.txt

# **Work To Be Done**
Fix the appropriate CMakeLists.txt 

# **Acceptance Criteria**
Code compiles what we expect it to based on the cmake flags
Flag does what it is supposed to do, which is toggle the building of the Utility programs on/off

spaulaus:
The flag is intended to control the compilation of these utility programs. We should update `Analysis/CMakeLists.txt` to actually use the flag properly.

Issue #251, opened 2017-03-06
Need to have variable TraceDelay


spaulaus:
# Description
I have identified two data sets (ISOLDE 2015 and ORNL 2016) that have trace delays that vary depending on the detector type that was used. The analysis software only supports a single trace delay. I have also been asked by users if this is something that we could support. Clearly there is a need. 

# Proposed solution(s)?
# Option 1
Set the trace delay via a similar interface to the WaveformRange or Fitting parameters. However, this would only provide control on a per detector type:subtype basis. At this level there's no guarantee that the type:subtype combination will all have the same delay (after all we assumed all types:subtypes had the same delay prior to this). This solution is pretty quick to implement since it just means we need to add a bit of code to the GlobalsXmlParser and the Globals class. 

# Option 2 
Create this as a channel parameter like a calibration. This will  ensure that we can set this parameter for each module/channel combination. This solution will take significantly more time to implement since we'll need a number of new classes to handle the new features. This also means that we should concatenate the WalkCorrector and Calibrator into a more comprehensive class (ChannelSettings?) **I think that this solution makes the most sense in the long term.**

tking53:
I think if we could quickly put Option 1 into effect, that would allow Option 2 to have the development time it needs to not be rushed. 
We could also during the implementation of Option 2 reimplement Option 1 as a "coarse" option. So we would have 3 ways to deal with this, Global (the way it is now), Coarse (Type:Subtype), and Fine (Channelwise). I also agree that Option 2 makes the most sense but it makes the Config File really long. (1 or 2 new lines per channel with traces; Im thinking of Dome-Vandle or NEXT ). So having ways to handle this on a coarser level than Channelwise, I think would be very useful.
Hopefully this is a niche feature, and we can continue as we always have. While having this as a backup if we do this by accident again.

spaulaus:
Well it's happened twice during experiments and a user has requested it. I don't think it's niche. What's most troubling is the experiment parts. What else have we gotten wrong?

tking53:
I have a xml type:subtype style fix for this. i will submit a PR after the holiday (and a bit more testing).

spaulaus:
Resolved this issue, see Pull Request #269

Issue #250, opened 2017-03-05
Need better error handling for legacy nodes in XML


spaulaus:
# Description
Many of the utkscan configuration files that are shipped with the software have not been updated to the latest XML scheme. The contain legacy nodes, and other formatting issues that will cause errors when they are used with the software. We should add some error handling to the parsing routines that checks for these legacy nodes and warns the user that they exist. Legacy nodes like the `Timing` node need to throw an exception when they're encountered because they will really throw a monkey wrench into the parsing of the Trace and Fitting nodes. 

# Work to be done
1. Identify all of the legacy nodes that are in the configuration files
2. Identify the nodes that could cause critical errors. 
3. Create a validation class (method?) that will validate nodes and look for the legacy nodes.
4. Update the example.xml so that it no longer contains legacy information. 

# Acceptance Criteria
1. All software compiles
2. Old XML files are parsed properly, and all legacy nodes are listed to the user
3. Old XML files throw errors if the legacy node encountered could cause a critical error. 
4. example.xml throws no errors.

spaulaus:
I'm updating the wiki as well.

PR #269, opened 2017-05-22
Resolving a number of issues

spaulaus:
The review process has become an impediment to some work. I have forked my own copy of the software so that I can quickly fix issues that are coupled.

# Resolved Issues
- #20
- #29
- #55
- #60
- #61
- #68
- #92
- #103 
- #105 
- #117
- #118
- #119
- #157 
- #228
- #233 
- #235
- #236 
- #237 
- #238
- #244
- #249
- #250 
- #251 
- #252
- #262 
- #271
- #272 
- #273 
- #274 
- #275 
- #276 

# Additional Updates
- RootScanner is now RootInterface. It lives in Resources and provides all programs an interface to a ROOT canvas.
- Reformatting code base to remove tabs, adhere to line lengths, and bringing things more in line with style guidelines.
- Corrected where `UtkUnpakcer::eventWidth_` is set so that it's set properly before any events are built.
- Removed QdcCompression from Globals and put it into the VandleProcessor constructor. This really shouldn't be used outside of the VandleProcessor anyway.


Analysis/Resources/include/ChannelConfiguration.hpp
```diff
@@ -0,0 +1,201 @@
...
+    ChannelConfiguration() : location_(9999), subtype_(""), type_("") {}
+
+    ~ChannelConfiguration() {}
+
+    ///Constructor taking arguments and setting values, this constructor is used to maintain functionality with the
+    /// previous Identifier class.
+    ///@param [in] atype : the type to set
+    ///@param [in] subType : the subType to set
+    ///@param [in] loc : the location to set
+    ChannelConfiguration(const std::string &atype, const std::string &subType, const unsigned int &loc) {
```
ksmith0: Rewrite to use initializer list.

Analysis/Resources/include/GslFitter.hpp
```diff
@@ -23,15 +23,6 @@ class GslFitter : public TimingDriver {
     ///Default Destructor
     ~GslFitter() {}
 
-    /// @return the amplitude from the GSL fit
```
ksmith0, line 26: Why did this get removed?

Analysis/Resources/include/ChannelConfiguration.hpp
spaulaus: Fixed in the next push

Analysis/Resources/include/GslFitter.hpp
spaulaus: These were moved to TimingDriver for enhanced compatibility.

spaulaus: ...and extend functionality.

Analysis/ScanLibraries/source/ScanInterface.cpp
```diff
@@ -978,26 +977,22 @@ bool ScanInterface::Setup(int argc, char *argv[],
         }
     }//while
 
-    //We check that the unpacker object has been set.
-    unpacker_ = unpacker;
-    if (!unpacker_)
+    if (!unpacker)
         throw invalid_argument("ScanInterface::Setup - The Unpacker object has not been set properly.");
+    else
```
ksmith0: You don't need the else here.

Analysis/ScanLibraries/source/ScanInterface.cpp
```diff
@@ -978,26 +977,22 @@ bool ScanInterface::Setup(int argc, char *argv[],
...
+        unpacker_ = unpacker;
```
ksmith0: You don't need the else here.

spaulaus:
> What if Minuit2 isn't installed? Can we do this or the gErrorLevel thing from hist if it is not?

I changed it so that if Minuit2 isn't returned from root-config then there's some pre-processor defines in there. That was simple and efficient way of doing this. The pre-processor define only covers a few lines of code, not whole swaths.

ksmith0:
That doesn't answer the issue with Minuit throwing up error messages everywhere. I believe histScanner was making use of `gErrorLevel` to suppress messages like this and we should be able to do the same with scope.

tking53:
Im curious why we wouldn't want the QDC compression available to other processors. Could it not be useful in any processor that deals with traces (any of the Beta or Liquid or maybe future HAGRiD processors), or are you thinking that we would have a different one per processor? 36f2201

tking53:
Also on behalf of Miguel and Robert I'm going to request this.

Can we have type:subtype defaults for the new Calibration nodes? They believe that the CFG is already too long.  
I see that the configuration default to Vandle:Medium settings but it might be nice to have the ability to set defaults for other type:subtype detectors with traces; like Vandle:Small or LaBr3 or whatever other large combined array robert comes up with in the future (VANDLE+NEXT has been floated around)

I believe that might ease the transition after this massive PR

https://github.com/spaulaus/paass/blob/5fb258d729f111d5cfe198aead033b89a0d49dda/Analysis/Resources/include/DefaultConfigurationValues.hpp#L10-L19

Analysis/Utilities/RootScanner/source/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+root_generate_dictionary(HistScannerDictionary HistScannerChanData.hpp
+        LINKDEF HistScannerLinkDef.h)
+add_executable(rootscan HistUnpacker.cpp HistScanner.cpp HistScannerChanData.cpp
+        HistScannerDictionary.cxx hist.cpp)
+target_link_libraries(rootscan PaassScanStatic ResourceStatic PugixmlStatic
+        PaassResourceStatic ${ROOT_LIBRARIES} -lTreePlayer)
```
ksmith0, line 6: Check this linking to TreePlayer.

ksmith0: ROOT suggests using the following:

```
find_package(ROOT REQUIRED COMPONENTS TreePlayer)
```

ksmith0 (approved):

PR #277, opened 2017-05-31
Resolves: Issue251 with type:subtype Trace Delays

tking53:
Allows us to have different type:subtype trace delays; This is the easy route for fixing #251 . This is needed for ORNL2016 data as M0 has a different delay than the rest of the modules. When calling GetTraceDelayInNS() you now can pass it a `type:subtype` string,  if it doesn't find a match in the map (or you don't pass it a string) it returns the global delay.  
It allows you to use the 'old' style (`<TraceDelay units="ns" value="264"/>`)  or the new style 
```
<TraceDelay units="ns">
         <global value="264"/> 
         <beta:double value="188"/>
         ...
</TraceDelay>
```
The global value is the only required subnode for the new style, the code will throw an error if it doesn't find it when parsing.
I have tested it with on kqxhc, as well as my Ubuntu 17.04 laptop. 
```
Linux Flavor = Ubuntu 17.04
Linux Kernel = 4.10.0-21-generic 
cmake version 3.7.2 
gcc (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406 
GSL version = 2.3
```

### DIFF
--- a/Analysis/Utkscan/analyzers/include/WaveformAnalyzer.hpp
+++ b/Analysis/Utkscan/analyzers/include/WaveformAnalyzer.hpp
@@ -33,6 +33,7 @@ public:
 
 private:
     Messenger *messenger_;//!< A pointer for the messenger class
+    std::string TDdet_; //!<string of type:subtype for the tracedelay function
 };
 
 #endif // __WAVEFORMANALYZER_HPP_
--- a/Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
+++ b/Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
@@ -44,10 +44,10 @@ void WaveformAnalyzer::Analyze(Trace &trace, const std::string &type,
 
     try {
         //First we calculate the position of the maximum.
+        TDdet_ = type + ":" + subtype;
         pair<unsigned int, double> max =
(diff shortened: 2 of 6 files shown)


spaulaus:
I fixed this in a more robust way on my branch. I recommend merging Pull Request #269.

ksmith0:
@tking53 Are you happy with the fix in #269? If so I suggest we close this pull request.

tking53:
I think we should move the base branch to RC after we merge dev over. since i think it would be helpful to have the option there; In case someone wants to rescan the ORNL2016 data quickly without updating the cfgs to the new channel wise style. I think we should have type:subtype defaults in the #269 version (if its not already there)

ksmith0:
What do you mean by base branch? Merge dev over to where?

tking53:
the base for the PR. and i was referring to after stan pulled dev into the RC; which it looks like he has already done. So i would close this PR and open a new one merging onto RC vs onto dev

spaulaus:
>  So i would close this PR and open a new one merging onto RC vs onto dev

I would hazard against anything getting ahead of dev. That is going to make merging dev into those branches tricky due to merge conflicts.

tking53:
yea, but i would really like the RC to have this as well as the PSPMT PR ( #278 ) and the MCAinPacman ( #221 which i merged earlier this afternoon) maybe we should revert the merge to RC till your ready with your PR?

PR #280, opened 2017-06-07
Release candidate into master

spaulaus:
I think it's time to update the master with all these bug fixes and enhancements. Do all the collaborators agree?


ksmith0:
I'm in favor of it, but I'm not using it everyday. I believe that the UT group should have a discussion about their current use patterns and if any issues had cropped up with the RC first.

spaulaus:
> I believe that the UT group should have a discussion about their current use patterns and if any issues had cropped up with the RC first.

I agree.

PR #282, opened 2017-06-08
#269 Do over

spaulaus:
Fixed and issue with utkscanor and fixed formatting on poll2.


spaulaus:
This PR is a do over of #269 . The first one wasn't rebased onto dev properly so things got kind of crazy. Those issues are now resolved.

PR #283, opened 2017-06-08
Removing a file that snuck in during the rebase

spaulaus:
This file has been duplicated and should only exist in Analysis/Scanor/include/


ksmith0 (approved):
This is in the wrong place.

PR #290, opened 2017-06-14
updated pspmt processor

akeeler:
PspmtProcessor updated to version used during the e14060 experiment


tking53 (approved):
It compiles and runs on my laptop (Ubuntu 17.04 root 6.08.06 gsl2.3)  and quack.

PR #296, opened 2017-06-22
Various new histos

tking53:
I had these to add before #269 / #282. 

Robert requested most of these during the IS632 experiment .

The list of new plots (in commit order):
1) Max Value of the trace vs beta detector (Even Left, Odd Right)
2) Always uncalibrated histograms for the VandleProcessor. These include Tdiff, BarNum vs ToF, and (per request ) QDC vs ToF. This last one is just to see that we are getting Something. These also all have "UNCALIBRATED:" in the histogram name; and are away from the usual ids (3300-3329)
3) Clover sums (2500 and 2510) without the cycling requirement. This is useful for cf or when there is a problem with the cycling logic.
4) Re-added the max value plot for the vandle bars and put the Average QDC vs (cor)ToF plots back to requiring a Time cal since they look very odd without one; and the need for always plotting these is alleviated by point 2

the last commit is just a small comment update in the DammPlotIds.hpp

### DIFF
--- a/Analysis/Utkscan/core/include/DammPlotIds.hpp
+++ b/Analysis/Utkscan/core/include/DammPlotIds.hpp
@@ -103,8 +103,8 @@ namespace dammIds {
 
     ///in GeProcessor.hpp
     namespace ge {
-        const int OFFSET = 2999;//!< Offset for CloverProcessor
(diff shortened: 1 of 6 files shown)


akeeler (approved):
builds and scans without issue

PR #297, opened 2017-06-23
added some global defaults

tking53:
1) Enabled global and module specific traceDelays in addition to the per channel ones, as well as reverted the units to ns (from samples). I do a conversion based on the frequency, this makes the value we put into Pixie  and whats in the config file the same base unit (seconds). This is how it was done before. I believe this will reduce some headache when making config files during experiments

2) Enabled Global Firmware and Frequency for use in nonmixed setups. 

I put both of these in the highest <Map> line; and the TraceDelay is required here, the Firm/Freq is only required to be set either here or in the module. The Module line overrides the global for that module, and similarly the per channel value overrides the module and global. This ensures that they are at least something for every module. 

I also changed when the MapNodeParser enters the channel node child parsers, so that the Trace and Fit child nodes are always set if the channel is a Vandle type since the defaults are for vandle:medium anyways. Again if it exists it should override the default behavior.

All of these changes make the transition to the new Dev style much easier, as well as keeping the config file from growing to monstrous  for full experiments. 

I have been using these for the last couple days without issue, with 17N data as well as the ORNL2016 data (which requires that the first module have a different Trace Delay than the rest)

### DIFF
--- a/Analysis/Resources/include/ChannelConfiguration.hpp
+++ b/Analysis/Resources/include/ChannelConfiguration.hpp
@@ -72,7 +72,7 @@ public:
     unsigned int GetTraceDelayInSamples() const { return traceDelayInSamples_; }
 
     ///@return The waveform bounds in samples
-    std::pair<unsigned int, unsigned int> GetWaveformBoundsInSamples() const { return waveformBoundsInSeconds_; }
+    std::pair<unsigned int, unsigned int> GetWaveformBoundsInSamples() const { return waveformBoundsInSamples_; }
 
     ///Check if an identifier has a tag
     ///@param [in] s : the tag to search for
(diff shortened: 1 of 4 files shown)


tking53:
here is what the beginning of my Map node looks like for ORNL2016 data; in case i wasn't clear in my explanation above.
```
    <Map verbose_calibration="False" verbose_map="False" verbose_walk="False" TraceDelay="264" firmware="34300" frequency="250">
      <Module number="0" TraceDelay="188">
        <Channel number="0" type="beta" subtype="double" tags="start,left">
          <Trace RangeLow="15" RangeHigh="5"/>
          <Fit beta="0.004265" gamma="0.1544"/>
        </Channel>
        <Channel number="1" type="beta" subtype="double" tags="start,right">
          <Trace RangeLow="15" RangeHigh="5"/>
          <Fit beta="0.004265" gamma="0.1544"/>
        </Channel>
```

akeeler (changes requested):
if map trace delay is not set, make error appear in a more accessible place.

akeeler (approved):

PR #298, opened 2017-06-23
adding Root Tree Output to VandleProcessor

tking53:
This adds root output to the VandleProcessor. @jeremybundgaard added this pre #269, and i brought it up. The tree seems to work like we thought. I wrapped it in a #ifdef; this is probably the wrong way to guard against errors when you don't have root installed.

The change to the DetectorDriverXmlParser.cpp is something i discovered when i turned `PAASS_USE_ROOT` off.


tking53:
closing since the branch has diverged. (both in what jeremy has and the dev branch)

PR #300, opened 2017-11-22
added methods to read external clock with pixie

jeremybundgaard:
Added methods (DecodeExternalTimeHigh,CalculateExternalTimeStamp) to XiaListModeDataDecoder Class to read external time stamp scalars from .ldf buffers and store them as data members of XiaData. Added methods: (S,G)etExternalTimeStamp + members:double externalTimeStamp_ to the XiaData Class


tking53:
changes here are included in #302. so im closing it

PR #302, opened 2019-02-05
Merge RIKEN2018 branch

tking53:
This branch contains the new features from the RIKEN2018 (ribf-136) experiment.

Summery of the major changes:
1) External Timing Clock-> added the ability to decode the external time stamp which is read in from the front pins.
1.5) updated poll2 commands to show the Ext Timing stuff (csr_test etc)
2)Root output from the detector driver. The various processors create a vector of detector events which is filled After the Process() call, once per pixie event. This means that root's Draw() from the interpreter will only work inside of a processor vector (i.e.  "vandle qdc vs vandle tof" is ok but "vandle tof vs clover energy" is not) if you want to make cross vector plots you need a script (like a TSelector) and the TTreeReader 
3) fixed a bug in the `utkscanor` initialization  which required a dummy `-o `argument. This is no longer required. 
4) new CloverFragProcessor which removes the more complex clover plots (most of which required MTC). This new Processor overlaps with the CloverProcessor in the DammPlotID space. so you can only use 1 or the other. They also fill the same vector in the root output 

Some other updates to a few of the Detector Processor. As well as refinement of the root output. 

This also includes the changes from #300, so i will be closing that one

As for testing, this has been my daily driver base for a while now. I have used it with ORNL2016 and E14060, RIBF136 data.

### DIFF
--- a/Acquisition/Interface/include/PixieSupport.h
+++ b/Acquisition/Interface/include/PixieSupport.h
@@ -112,10 +112,10 @@ public:
 
     BitFlipper() {
         bit = 0;
-        num_toggle_bits = 19;
+        num_toggle_bits = 22;
     }
 
-    BitFlipper(unsigned int bit_, unsigned int num_toggle_bits_ = 19) {
+    BitFlipper(unsigned int bit_, unsigned int num_toggle_bits_ = 22) {
         bit = bit_;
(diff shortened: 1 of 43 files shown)


rin-yokoyama (approved):
It has been works since the experiment. It is stable.

PR #304, opened 2019-02-05
Fixes issue 303

### LINKED ISSUE #303: Place Cycle always asked for
This is related to the FillLogicStruc() method. for the root output (fixing now so short and sweet)

tking53:
Forgot to protect the FillLogicStruc() from negative cases, where we dont have logic signals

Fixes #303


pbrionnet (approved):
It works

PR #305, opened 2019-05-16
2019updates

tking53:
Lots of upgrades and bug fixes. 

Most important changes are (in no particular order) : 
1) GetTime() changed to ALWAYS return ADC clock ticks and GetTimeSansCFD() always returns DSP clock ticks. 
     - Meaning that the conversion for GetTime() is ALWAY GetAdcClockInSeconds() and GetTimeSansCfd() is ALWAYS either GetClockInSeconds() or GetFilterClockInSeconds()

2) VANDLE
      - Geometry corrected Time of Flight (corTof) for VANDLE now corrects all bars to a single distance.
      -  Fixed the SpeedOfLightIn...Bar constants to always be set. Because this is set by the GlobalsXmlParser.cpp the easy way is to just always set them, because we cant depend on the Map or the DetectorDriver nodes being parsed already. We retain the ability to override them in this node (any combination of them)

3) Multi Crate related:
      - New overloads for GetClockInSeconds(), GetAdcClock..., GetFilterClock... that take the module frequency as input to return the correct tick -> second conversion factor. (Only available with Rev F set in the <Global> node of the xml)
      - Added module frequency to the ChanID class.

4) New processor for detector development and testing (RootDevProcessor). This processor tries to dump as much of the information from the ChanEvent as possible. 

      - Raw Energy
      - Calibrated Energy
      - Evt Time Sans Cfd in ns
      - Evt Time in ns
      - Location in the map
      - Module number
      - Channel number 
      - Subtype 
      - Group
      - Pileup (from pixie)
      - Saturation (from pixie)
      - Trace (if present)
      - Position of the Max (if we have a trace)
      - Value of the Max (if we have a trace)
      - Trace QDC (if we have a trace)
      - High Res Time in ns (if we have a trace)
      - The onboard QDC sums if present in the list mode

5) Added initial support for the new R42159 firmware with its new CCSRA bits 

6) Added some trace based analysis to the PspmtProcessor, to reject events with traces which are obviously bad, reflections, induced signals from ion implantation in the high gain channels, etc. These tests require the Waveform Analyzer but not the Fitting Analyzer. The Waveform Analyzer ignore list works like before, but the Fitting Analyzer's lets the Waveform Analyzer do its thing then stop before starting the fit.

7) Marked some of the cmake options that we don't use as Advanced to clear up the ccmake output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ install/
 *.layout
 *.save
 .idea/
+.vscode
 
 #DAMM files
 *.drr
--- a/Acquisition/Interface/include/PixieSupport.h
+++ b/Acquisition/Interface/include/PixieSupport.h
@@ -112,10 +112,10 @@ public:
 
     BitFlipper() {
         bit = 0;
-        num_toggle_bits = 22;
+        num_toggle_bits = 24;
     }
 
-    BitFlipper(unsigned int bit_, unsigned int num_toggle_bits_ = 22) {
+    BitFlipper(unsigned int bit_, unsigned int num_toggle_bits_ = 24) {
         bit = bit_;
         num_toggle_bits = num_toggle_bits_;
     }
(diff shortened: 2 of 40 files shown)


rin-yokoyama (approved):
It works.

PR #308, opened 2020-07-05
Pull in the updates from the mix_bit_upgrades branch

tking53:
This branch includes upgrades stemming from the work testing the mixed bit same frequency capabilities of utkscan(or).

This branch has been in use with multiple analyses, and are considered stable enough for merging into dev.

Highlights include:
Protections for the poll2 command "toggle". Previously if you gave an invalid bit name (polarity, gain, traces, etc.) toggle would flip bit 0. This behavior has been fixed by having the name -> bit list check fail and output an error before the call to the API.

We also added some safety checks to toggle, adjust_offsets, pwrite and pmwrite. For whatever reason the API lacks sanity checks for these functions and will let you try to run them on a module that is not installed. (a 3 module system would let you write to module 3 rather than just 0,1,2) this would cause the computer to Hard lock and require a power button forced reboot. pread and pmread had protections in the API but i have added our checks as well for consistency's sake. 

We have added support for the 42159 and 42950 ( the newest vandle firmwares 12b, and 16b respectively) 
We have also added support for parsing the 500MHz general firmware. Right now this is Hard matched since we only have 1. This will be fixed once we have more firmwares, and the parsing functions have been updated to reflect the differences in the list more format (specifically Header Word 2 which is structured differently since the "forced trigger bit is now 3 bits )

We have made Ctrl+c clear the line in poll2, rather than just spitting out an error saying we blocked the keystroke 
We have also fixed the INS key behavior, previously you could enter "insert"/"overwrite" mode but were unable to exit it. 

We have given the CfdAnalyzer some of the upgrades that were previously given to the Fitting and Waveform analyzers. Namly the ignore list.

We fixed the MapNodeXmlParser's parsing of the Cfd Parameters, and we have made the parser ALWAYs parse the cfd node for the "vandle" type, which is also done for the Fit and Trace nodes.

Finally the GeProcessor has gotten root output support. It is filled into the clover_vec_  with cloverNum = -1, since detNum is type:subtype specific. This will let us use the two processors together and identify them later in the root analysis.

### DIFF
--- a/Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
+++ b/Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
@@ -25,9 +25,17 @@ WaveformAnalyzer::WaveformAnalyzer(const std::set<std::string> &ignoredTypes)
     extremeBaselineRejectCounter_ = 0;
 }
 
+bool WaveformAnalyzer::IsIgnoredDetector(const ChannelConfiguration &id) {
+    if (IsIgnored(ignoredTypes_, id)) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
 void WaveformAnalyzer::Analyze(Trace &trace, const ChannelConfiguration &cfg) {
     TraceAnalyzer::Analyze(trace, cfg);
-    if (trace.IsSaturated() || trace.empty() || ignoredTypes_.find(cfg.GetType()) != ignoredTypes_.end() || ignoredTypes_.find(cfg.GetType() + ":" + cfg.GetSubtype()) != ignoredTypes_.end() || ignoredTypes_.find(cfg.GetType() + ":" + cfg.GetSubtype() + ":" + cfg.GetGroup()) != ignoredTypes_.end()) {
+    if (trace.IsSaturated() || trace.empty() || IsIgnored(ignoredTypes_,cfg)) {
         trace.SetHasValidWaveformAnalysis(false);
         EndAnalyze();
         return;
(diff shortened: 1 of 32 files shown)


Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
```diff
@@ -64,14 +73,26 @@ void WaveformAnalyzer::Analyze(Trace &trace, const ChannelConfiguration &cfg) {
...
         // hard limit of 50 as a cutoff since anything with a standard
         // deviation of this high will never be something we want to analyze.
-        static const double extremeBaselineVariation = 50;
-        if (baseline.second >= extremeBaselineVariation) {
+        //We are switching to a percentage of the baseline value, because higher 
+        // bit resolution Pixies are able to capture more variation in the baseline.
+        //Also if the avg baseline is lower than 10 ADC units (for any bit resolution),
+        // this is also a sign of a bad trace capture.
+        
+        static const double extremeBaselineVariation = 0.15 * baseline.first; //Checking for an std of 15% of the avg baseline (this way we are sensitive to the different bit resolutions)
```
rin-yokoyama: We should remove static keyword here. Otherwise the extremeBaselineVariation is fixed to the value in the first call which I think is not intended.

rin-yokoyama (changes requested):

Analysis/Utkscan/analyzers/source/WaveformAnalyzer.cpp
tking53: yes your right.

tking53: your right.

rin-yokoyama (approved):
It now looks good to merge

PR #315, opened 2023-11-17
Merge Over the FDSi year one updates and upgrades into dev

tking53:
This is a combined merge of the fdsi and fdsi-mtas branches (they were combined outside of GH and tested). This has been used and tested over the last year.

there are 2 commits at the top of the tree which are new. These fix a bug with the root output which caused it to not write the last buffered set of events from the TFile write cache. This was caused by us not properly handling the program exiting sequence from the fortran components. This was fixed by simply invoking the Deconstructors before calling CleanUp() on the fortran codes. 

the second commit is a small change to the TFile write options so that we dont have duplicate TKeys in the root file once its closed. There was only one set of data but multiple copies of the metadata, this small change just tells ROOT to write the updated metadata and then delete the old set of metadata during the TTree::Write() calls.


Analysis/Utkscan/core/include/SegmentDetector.hpp
```diff
@@ -0,0 +1,119 @@
...
+    };
+    virtual void SetSegBack(ChanEvent* a) {
+        segBack_ = a;
+    };
+
+    virtual void SetPixieRev(string &a) {
+        PixieRev = a;
+    };
+
+    virtual string GetPixieRev(){
```
jmattspartacus: Seems like this should probably be const noexcept

Analysis/Utkscan/core/include/SegmentDetector.hpp
```diff
@@ -0,0 +1,119 @@
...
+        return PixieRev;
+    };
+
+    virtual bool IsValidSegment() const {
```
jmattspartacus: Should probably be noexcept, unless you can forsee a situation where we'd possibly be accessing nullptr or other badness here.

Analysis/Utkscan/processors/include/MtasImplantSipmProcessor.hpp
```diff
@@ -0,0 +1,64 @@
...
+    /** Declares the plots for the class */
+    void DeclarePlots(void);
+
+    double CalOnboardQDC(int bkg, int waveform, const std::vector<unsigned> oqdcs, double bkgTimeScaling = 1, double waveformTimeScaling = 1);
+
+    std::pair<int, int> ComputeSiPmPixelLoc(int xmlLocation_);
+
+    void FillRootStruct(ChanEvent* evt, double& onboardqdc, const std::pair<int,int> &positions = {0,0});
+
+    std::pair<double,double> CalculatePosition(std::vector<std::vector<double>>& data);
```
jmattspartacus: Can probably mark const

Analysis/Utkscan/processors/source/MtasProcessor.cpp
```diff
@@ -0,0 +1,411 @@
...
+        //! begin loop over segments for sums
+	double centerSum = 0;
+	double innerSum = 0;
+	double middleSum = 0;
+	double outerSum = 0;
+	double totalSum = 0;
+	//!USE LAMBDA FUNCTION FOR THIS. MAKES TYPOS LESS LIKELY
+	//auto energyfunc = [](double a, double b) { return (a + b) / 2.0; };
+	//auto tdiffFunc = [](double a, double b) { return a - b; };
+	for (auto segIter = MtasSegVec.begin(); segIter != MtasSegVec.end(); ++segIter) {
```
jmattspartacus, line 233: Why not `for(auto segIter: MtasSegVec){` instead?

Analysis/Utkscan/processors/source/MtasProcessor.cpp
```diff
@@ -0,0 +1,411 @@
...
+			totalSum += segmentAvg;
+			middleSum += segmentAvg;
+		} else if (segmentID >= 18 && segmentID <= 23 && segIter->IsValidSegment() ){
+			double segmentAvg = segIter->GetSegAvgEnergy();
+			totalSum += segmentAvg;
+			outerSum += segmentAvg;
+		}
+	}
+	//! loop over segments for ploting
+	for (auto segIter = MtasSegVec.begin(); segIter != MtasSegVec.end(); ++segIter) {
```
jmattspartacus, line 254: Why not for(auto segIter: MtasSegVec){ instead?

Analysis/Utkscan/processors/source/PidProcessor.cpp
```diff
@@ -27,325 +27,896 @@ using namespace std;
...
+      if((*pinLogic))
+        pid_struct.cross_pin_0_logic_time = (*pinLogic)->GetTime() * internalTAC_Convert_Tick_adc;
+   }
+   if(!b2Logic_vec.empty() ){
+      auto b2Logic = std::max_element(b2Logic_vec.begin(), b2Logic_vec.end(), compare_energy);
+      double internalTAC_Convert_Tick_adc = Globals::get()->GetAdcClockInSeconds((*b2Logic)->GetChanID().GetModFreq()) * 1e9;
+      if((*b2Logic))
+        pid_struct.cross_scint_b2_logic_time = (*b2Logic)->GetTime() * internalTAC_Convert_Tick_adc;
+   }
+   //std::cout<<"come to line 251"<<std::endl;
```
jmattspartacus, line 277: Debug output like this should be removed, lines change each time code is added/deleted.

Analysis/Utkscan/processors/source/PidProcessor.cpp
```diff
@@ -27,325 +27,896 @@ using namespace std;
...
+         }
+      }
+   }
+   //std::cout<<"come to line 344"<<std::endl;
+   //
+   //! TOF 2 group (image L to cross scint onboard cfd)
+   if (!imageL_vec.empty() && !b2Logic_vec.empty()) {
+      // Get elements with the largest energy in this event for image L
+      auto imageL = std::max_element(imageL_vec.begin(), imageL_vec.end(), compare_energy);
+      double internalTAC_Convert_Tick_adc_imageL = Globals::get()->GetAdcClockInSeconds((*imageL)->GetChanID().GetModFreq()) * 1e9;
```
jmattspartacus, line 369: This will segfault if imageL is nullptr. See comment at the start of PidProcessor::PreProcess

Analysis/Utkscan/processors/source/PidProcessor.cpp
```diff
@@ -27,325 +27,896 @@ using namespace std;
...
+
+      // Get elements with the largest energy in this event for pinLogic
+      auto b2Logic = std::max_element(b2Logic_vec.begin(), b2Logic_vec.end(), compare_energy);
+      double internalTAC_Convert_Tick_adc_b2Logic = Globals::get()->GetAdcClockInSeconds((*b2Logic)->GetChanID().GetModFreq()) * 1e9;
```
jmattspartacus, line 373: This will segfault if b2Logic is nullptr. See comment at the start of PidProcessor::PreProcess

Analysis/Utkscan/processors/source/PidProcessor.cpp
```diff
@@ -27,325 +27,896 @@ using namespace std;
...
+      // Check for nullptr
+      if ((*imageL) && (*b2Logic)) {
```
jmattspartacus, line 375: Check is redundant since we will have already segfaulted by here if either is nullptr. See comment at the start of PidProcessor::PreProcess

Analysis/Utkscan/processors/source/PidProcessor.cpp
```diff
@@ -27,325 +27,896 @@ using namespace std;
...
+   rit_threshold_ = RIT_thresh;
 
-    yso_threshold_ = YSO_Implant_thresh;
-    fit_threshold_ = FIT_thresh;
+   yso_energy_prev = 0;
+   fit_energy_prev = 0;
+   fit_energy_prev = 0;
 }
 
 bool PidProcessor::PreProcess(RawEvent &event) {
```
jmattspartacus, line 200: Are we really worried about imageL/R, dispL/R and such being nullptr in this function? If this is a concern, then each time we grab them from the vectors, for instance in this block
```cpp
if(!imageL_vec.empty() ){
  auto imageL = std::max_element(imageL_vec.begin(), imageL_vec.end(), compare_energy);
  double internalTAC_Convert_Tick_adc = Globals::get()->GetAdcClockInSeconds((*imageL)->GetChanID().GetModFreq()) * 1e9;
  if((*imageL)){
    pid_struct.image_scint_L_logic_time = (*imageL)->GetTime() * internalTAC_Convert_Tick_adc;
  }
}
```
we need to put anything using (*image) behind a guard like I've done. It seems like the checks for nullptr might not be neccessary, based on the code getting this far.
```cpp
if(!imageL_vec.empty() ){
  auto imageL = std::max_element(imageL_vec.begin(), imageL_vec.end(), compare_energy);
  if((*imageL)){
    double internalTAC_Convert_Tick_adc = Globals::get()->GetAdcClockInSeconds((*imageL)->GetChanID().GetModFreq()) * 1e9;
    pid_struct.image_scint_L_logic_time = (*imageL)->GetTime() * internalTAC_Convert_Tick_adc;
  }
}
```

Analysis/Utkscan/processors/include/PspmtProcessor.hpp
```diff
@@ -76,7 +76,7 @@ class PspmtProcessor : public EventProcessor {
     /// will use to calculate the position.
     ///@return The x,y position of the interaction
     std::pair<double, double> CalculatePosition(double &xa, double &xb, double &ya,
```
jmattspartacus: Can mark const and possibly mark noexcept if we can guarantee that none of the combinations of the denominators will ever be 0

Resources/include/StringManipulationFunctions.hpp
```diff
@@ -12,6 +12,15 @@
 #include <vector>
 
 namespace StringManipulation {
+    ///Converts full string to lower case. 
+    /// @param [in] s : String to make lowercase
+    /// @return the lowercase string
+    inline std::string StringLower(std::string s){
+        std::transform(s.begin(),s.end(),s.begin(),::tolower);
```
jmattspartacus, line 19: Maybe add std to the :: here? Not really necessary, just nitpicky

jmattspartacus (changes requested):
The only place I think we absolutely need to make sure we clean up/take care of before merging is PidProcessor::PreProcess, there are a lot of places where there are guards for nullptr, but a ptr to the object is already used, which would segfault.
Everything else is just nitpicks.

Analysis/Utkscan/processors/source/CloverFragProcessor.cpp
```diff
@@ -173,7 +173,7 @@ bool CloverFragProcessor::Process(RawEvent &event) {
         }
 
         if (DetectorDriver::get()->GetSysRootOutput()) {
-            //Fill root struct and push back on to vector
+            //Fill root struct and push back on to vector3
```
tking53: There is also a typo here. Making a note so I remember to fix it

Analysis/Utkscan/processors/source/MtasImplantSipmProcessor.cpp
```diff
@@ -0,0 +1,387 @@
+/** @file MtasProcessor.cpp
+ * @brief  Basic MtasProcessor for MTAS at FRIB
```
tking53: Doxygen tags need to be fixed for the MTAS sipm processor

Analysis/Utkscan/processors/source/MtasImplantSipmProcessor.cpp
```diff
@@ -0,0 +1,387 @@
...
+const unsigned int DD_DY_L_TQDC = 8;
+const unsigned int DD_DY_H_ENERGY = 9;
+const unsigned int DD_DY_H_OQDC = 10;
+const unsigned int DD_DY_H_TQDC = 11;
+const unsigned int DD_SIPM_PIXEL_IMAGE_HG = 12;
+const unsigned int DD_SIPM_PIXEL_IMAGE_LG = 13;
+const unsigned int DD_SIPM_HIRES_IMAGE_LG = 14;
+const unsigned int DD_SIPM_HIRES_IMAGE_HG = 15;
+const unsigned int DD_SIPM_HIRES_IMAGE_LG_QDC = 16;
+const unsigned int DD_SIPM_HIRES_IMAGE_HG_QDC = 17;
```
tking53, line 44: Given the recent confusion about which qdc we mean here, do we want to add trace qdc based images?

Analysis/Utkscan/processors/source/MtasProcessor.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+		bool isFront = (*chanEvtIter)->GetChanID().HasTag("front");
+		bool isBack = (*chanEvtIter)->GetChanID().HasTag("back");
+		int chanOffset;
+		if (isFront) {
+			chanOffset = 0;
+		} else if (isBack) {
+			chanOffset = 1;
+		} else if ((isFront && isBack) || (!isBack && !isFront) ) {
+			cout<<"ERROR::MtasProcessor:PreProcess BOTHESSSSS ("<<(*chanEvtIter)->GetModuleNumber()<<" , " << (*chanEvtIter)->GetChannelNumber() << ") !"<<endl;
```
tking53: Probably need a more helpful error here

Analysis/ScanLibraries/source/XiaListModeDataDecoder.cpp
```diff
@@ -227,7 +227,8 @@ std::pair<unsigned int, unsigned int> XiaListModeDataDecoder::DecodeWordZero(con
 
     data.SetChannelNumber(word & mask.GetChannelNumberMask().first);
     data.SetSlotNumber((word & mask.GetSlotIdMask().first) >> mask.GetSlotIdMask().second);
-    data.SetCrateNumber((word & mask.GetCrateIdMask().first) >> mask.GetCrateIdMask().second);
+    // Crate number in Pixie list-mode data is ignored
+    data.SetCrateNumber(0);
```
tking53, line 231: We need to ask nori why he does/wants this? The crate number should be in the list data, if it's not that's a bug in the firmware. If it is why doesn't he want to track it?

Analysis/Utkscan/CMakeLists.txt
```diff
@@ -8,11 +8,13 @@ option(PAASS_UTKSCAN_GAMMA_GATES "Gamma-Gamma gates in the Gamma-Ray Processors"
 option(PAASS_UTKSCAN_ONLINE "Options for online scans" OFF)
 option(PAASS_UTKSCAN_TREE_DEBUG "Debugging info for TreeCorrelator" OFF)
 option(PAASS_UTKSCAN_VERBOSE "Make Scan More Verbose" OFF)
+option(PAASS_OVERRIDE_EXE_PREFIX "Override the scan exe prefix" OFF)
```
tking53, line 11: These upgrades should be very loud at the end of the compilation (or install) step to alert the user that the default has changed. To avoid a situation where someone sets this up and has other users who might need to recompile and forget or don't know that the executable's name has changed

tking53 (commented):
I added some comments on some typos I would like to fix (but somehow missed before)

tking53:
> The only place I think we absolutely need to make sure we clean up/take care of before merging is PidProcessor::PreProcess, there are a lot of places where there are guards for nullptr, but a ptr to the object is already used, which would segfault.
> Everything else is just nitpicks.

I think the lines which use the unguarded ptrs probably came during the experiment in a rush. We should definitely decide if we need the check for nullptr or if !empty() is enough.

Analysis/Utkscan/processors/include/MtasImplantSipmProcessor.hpp
truland: Can't mark FillRootStruct const as it modifies the struct that is pushed onto the vector.

Can mark CalculatePosition as const as well as the vector passed into it by ref to prevent modification.

Analysis/Utkscan/core/include/SegmentDetector.hpp
truland: completed

Analysis/Utkscan/core/include/SegmentDetector.hpp
truland: completed

Analysis/Utkscan/processors/include/PspmtProcessor.hpp
truland: Fixed const correctness, not marking noexcept because the check for zero denom is not easy and is better to just eat the NaN/Inf we would get out of divide by zero

Resources/include/StringManipulationFunctions.hpp
truland: Can't use std::tolower without passing in a lambda function, and to support older compilers don't want to make that possibly breaking change

Analysis/Utkscan/processors/source/PidProcessor.cpp
truland: GetSummary can return a null DetectorSummary if the object doesn't exist already. So the calls significantly earlier when we do `event.GetSummary("pid:cross_scint_b1", true)->GetList();`, but the true second argument causes the summary to be constructed assuming the xmldriver hasn't already made the summary. But yeah according to DetectorDriver.cpp, if the list isn't empty then the objects are never null. This will take a while to fix appropriately.

truland: Should just merge it as it is now as the check is basically free and this should be addressed in a different PR/issue

Analysis/Utkscan/processors/source/CloverFragProcessor.cpp
truland: fixed

Analysis/Utkscan/processors/source/MtasImplantSipmProcessor.cpp
truland: Should be addressed in a different PR/issue as that requires input on how to generate the trace QDC

tking53: The trace qdc comes from the waveform analyzer. We would need to duplicate the image vectors and some extra math. Not a big deal and is easy to punt to a subsequent PR. Especially since it isn't a workable mode for actual experiments (trace data rates are too much)

Analysis/Utkscan/CMakeLists.txt
truland: doneish

jmattspartacus (approved):

PR #316, opened 2023-12-07
Changes plots to use array storage, use write thread for his file output

jmattspartacus:
Changes plots and drr_entry storage to use vectors instead of maps, saves some time.
Implements a way of doing sequential writes to disk for filling his files. The writes are done in a second thread, after having the writes sorted by buffer location using an ordered map.

See here for a history of the relevant commits.
https://github.com/tking53/paass/pull/4


tking53 (approved):
Sorry for the delay.  This works quite well.  It speeds up the scan by a significant factor.  In the below image, the "james_runtime_txt*" files are the 5 scans of an EVT file from nscldaq using this PR (4b87c3), and the  "runtime_txt" is pixie16/dev (@8408ce). These scans were run sequentially (5 of one then the 5 of the other) on donda @ UTk, and was written to the HOME raid, no precautions were taken against other users occupying that RAIDs bandwidth. 

image

PR #317, opened 2024-01-15
ScanInterface:: Fix to utkscan segfaulting with ldfs on high chunk count spills

tking53:
For example, FDSi MTAS SiPm implant with forced anode readout or FDSi VANDLE in doubles with a hot source and low thresholds. We were not allocating enough memory to read the data into. This was NOT an issue with scanor because it just reads 8192 words at a time (rather than a full spill) and passes it up as chunks. This will need a more permanent fix with a vector at a later date. Thanks to T. Ruland for finding and fixing this.


truland (approved):
Of course I approve, I helped write it.
This fixes many things on several systems, a similar fix is needed for the shm as well, but that is unimportant.

PR #318, opened 2024-02-09
New Monitor programs.

tking53:
This PR includes the new main and sub monitor programs. 
The mainmonitor rebroadcasts the information received from poll2 onto a maximum of six submonitor sockets. 
The new monitors both support color coded output and multiple rows. There are different color coding groups as well (to be updated to have appropriate ranges).


Acquisition/Poll/source/mainmonitor.cpp
```diff
@@ -0,0 +1,271 @@
...
+    for (int i = 0; i < MAX_NUM_SUBMONITORS; ++i) {
+        Submonitor_Clients.at(i).SendMessage(buffer, msg_size);
+    }
+}
+
+void mainmonitor::OpenSubmonitorSockets() {
+    std::cout << " Opening relay ports for submonitors" << std::endl;
+    std::string submonitorPortMessage = " Opened Client on ports ";
+    for (int i = 0; i < MAX_NUM_SUBMONITORS; ++i) {
+        Submonitor_Clients.emplace_back(Client());
```
jmattspartacus, line 54: Since Client has a parameterless constructor, using emplace_back(Client()) here implicitly uses a copy constructor.

truland: emplace_back only has move call signs, the destructor will be called on the r-value temporary when it gets moved into the vector. 
All will be fine since how little the monitor program actually does and since we limit the number of submonitors statically within the source code and not passed in as a command line parameter the accidental 6 extra allocations will be fine.
This does make it do extra work, but the vector takes ownership of the object after the emplace_back.
When the vector is destroyed at close of program the memory will be freed still.

jmattspartacus: I just wanted to point this out, it's not an issue for the functionality, just a nitpick on my part.

tking53:
@truland has ok'd the MTAS color ranges. So i think this is ready to do now.

truland (approved):
yep, sure does seem to work. this will be ian's fault if it doesn't

PR #320, opened 2024-07-18
Adding Mixed Frequency event building and RIBF168 things 

tking53:
## New Features:
1) We now build events in time space rather than tick space which allows us to have crates with mixed frequency setups. 
   - This was tested with pulsers, and then at RIBF168.
   - This required the removal of the "Globals::get()->GetClockInSeconds()" family of functions.
   - We now do the conversion to time when actually unpacking and decoding the binary data.
   - The replacement is two new functions "XiaData::GetTimeInNs()" and "XiaData::GetTimeSansCfdInNs()"
      - Processors have been updated to utilize the new functions.
   -  The XML now requires the "revision" to be in the <Map> node rather than the <Global> node. This follows the same logic as the Global Trace Delays, firmware, frequency (which was already there). This is required because with the addition of the RevHs the 250MHz and 500MHz boards will now have two possible conversion factors from the standard EVT timestamp to nanoseconds (different base clocks of the FPGAs). Wiki Updates for this are pending. 
2) New PSPMT Summary information added to ROOT output.
   - This was an addition requested and initially authored by Ian Cox (UTk). This allows us to do some rudimentary ion beta correlations without the full "merger" pipeline.
   - It also allows for some rudimentary beam diagnostics for FDSi ( or other experiments where we have access to the full PID suite in the Pixie data stream)
3) Added two new "pwrite" parameters used by the General firmware for validation trigger generation. To my knowledge these have no effect when using the VANDLE firmware.  
4) New RIKEN Specific PID Processor
    - TO avoid guting the normal PIDProcssor which has become tuned for FRIB and FDSi we wrote a new simpler one. 
    - My plan is to rename the PidProcessor to the FDSiPidProcessor at some point in the future and possibly clean/split it into an FP1 and FP2 variants for easier readability/maintenance. But this is to be decided. 
 5) New POLL2 command variants
     -  For historical and muscle memory reasons, I have added "htit", "ouf" as requested by T. Ruland
 6) RootDevProcessor can now take extra types defined in the XML.
     - Useful addition, which doesnt require significant changes to the XML if someone wants/needs the RootDev output for a specific type/ detector
     - Added the external timestamp to the RootDev Output. Mostly for completeness but was used for confirming that this feature existed in the General Firmware which was to be used with RIBF168.
  7) New UTKSCAN / UTKSCANOR command line flag "p"
     - This flag allows for the customization of the printing frequency of the "Build X events in Y seconds" messages. This are printed while the scan is progressing, but does cause some slow downs in the process. This allows one to decrease (or increase) the printing frequency from the standard 5000 to a user desired value.
  8) Basic automatic elog posting in non-pacman poll2
      - This currently works and prints to an elog the current file prefix, the run start / stop time, the title and some other details to a run start and run stop posts. 
      - This is currently not as complete as I would like it, and is not documented basically at all. This will be coming to the wiki soon. Contact T. T. King (kingtt at ornl dot gov) for assistance in the meantime. 
           - Mostly i do not like where certain settings are stored and read from ( elog address, port and password mostly) 
           - I want to get this merged over since this is in the middle of this long set of work.
  9) Proper decoding of the 500MHz onboard CFD information.
     - The 500MHz CFD has some specifics which required changes to how we store some of the binary information. 
     - Since the "forced bit" is now a value of the "source bit" we now check for that explicitly and set "forced".
     - Tested with a pulser, I get reasonable values for things and they are bounded by reality. 
     - Fixed the "multiplier" between CFD units and EVT units from 10 -> to the correct 5. This was noted by me in 2019 while looking at other things, but i did not have a 500 to test so i had just noted it.
     - Ian had hot fixed this bug away by tweaking the GammaScintFragProcessor (since that was the processor he was trying to use) I revered that commit after fixing the underlying issue.


truland (approved):
Tested by Ian, his fault if it isn't working.

PR #321, opened 2024-08-08
XiaListModeDataDecoder::CalculateTimeInSamples() Fix for high res onboard timing

tking53:
Forgot the `break` statements for the switch statement. This was causing fall through, that wasn't caught due to a short term lack of non 500MHz digitizers.


truland (approved):
I blame Ian and Toby for telling me it works last time.

jmattspartacus (approved):

